Issue #23, opened 2017-05-09
Write a test helper


meetmangukiya:

meetmangukiya:
Ways in which I can test plugin:
# Way 1

```py

# plugin file

class Plugin(BotPlugin):

  @staticmethod  
  def helper_func():
    # ... 
    
  @botcmd
  def cmd(seld, message, args):
    self.helper_func()

# test file

import plugin

def test_plugin():
  assert plugin.helper_func() == <assertion value>
```

Udayan12167:
I would prefer the second method as it is the intended errbot testing method I think.

Udayan12167:
Would help us avoid some errors.

meetmangukiya:
Second method has been used in five of plugins and all seem to work well. I think I'm gonna stick with second way

meetmangukiya:
(Closing)

PR #51, opened 2017-05-24
Plugin: Nevermind

meetmangukiya:


plugins/nevermind.py
```diff
@@ -0,0 +1,14 @@
+from errbot import BotPlugin, re_botcmd
+
+class Nevermind(BotPlugin):
+    """
+    Doesn't mind
+    """
+
+    @staticmethod
+    def nm_msg():
+        return "I'm sorry :("
```
SanketDG: Do we really need a separate function? :smile:

tests/nevermind_test.py
```diff
@@ -0,0 +1,12 @@
...
+
+from plugins.nevermind import Nevermind
+
+pytest_plugins = ['errbot.backends.test']
+
+extra_plugin_dir = '../plugins'
+
+def test_nevermind(testbot):
+    testbot.push_message('!nevermind')
+    assert "I'm sorry :(" in testbot.pop_message()
```
SanketDG: would it be possible to use `pytest` here?

tests/nevermind_test.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+from plugins.nevermind import Nevermind
+
+pytest_plugins = ['errbot.backends.test']
+
+extra_plugin_dir = '../plugins'
```
SanketDG: what does this do?

plugins/nevermind.py
```diff
@@ -0,0 +1,14 @@
+from errbot import BotPlugin, re_botcmd
+
+class Nevermind(BotPlugin):
+    """
+    Doesn't mind
```
SanketDG: This is redundant information from the `.plug` file, can you check upstream to see a way around this? Maintaining two sets of docs is  hard.

plugins/nevermind.py
meetmangukiya: I am still finding best way to test things.. 

There are two options so far:
1. Use the pytest plugin that errbot provides
2. Form staticmethods and test the output of those static methods. Instantiating the Plugin is not possible since for initialization we need the bot instance itself, which is what is done in the pytest plugin

tests/nevermind_test.py
meetmangukiya: What do you mean :/ ? It is the pytest plugin that passes the testbot in the first place.

tests/nevermind_test.py
meetmangukiya: Load the custom plugins from the given directory

plugins/nevermind.py
meetmangukiya: Sure, I'll check with upstream

tests/nevermind_test.py
SanketDG: Could you add a comment explaining so?

tests/nevermind_test.py
SanketDG: Sorry, I was thinking of a class based test suite, but it looks overkill, let's drop it.

plugins/nevermind.py
SanketDG: If we are doing everything through the static methods, then its perfectly fine to only test them.

Also, testing both the command and the helper method isn't worth it imo.

meetmangukiya: I created an issue, please see https://github.com/coala/corobo/issues/23#issuecomment-303944269 

It is in discuss mode, please leave a feedback there :D

meetmangukiya: @SanketDG yeah, I think we have to decide whether we just want to test the helper functions or do a full fledged unit as well as integration test with errbot as well.

meetmangukiya:
@SanketDG PTAL :D

plugins/nevermind.py
meetmangukiya: The documentation is plug file is external documentation, it is not used in the help command. The documentation that is used in help command is the docstrings, they can be fairly independent and fwiw we don't have to maintain documentation of `plug` file

meetmangukiya:
@SanketDG I am merging this, I need a plugin in master

meetmangukiya:
@rultor merge

PR #52, opened 2017-05-25
Plugin: The rules

meetmangukiya:


Udayan12167:
Can you package both files in a folder inside the plugins folder. Further plugins might have multiple files. It would be better to package that way.

plugins/the_rules.py
```diff
@@ -0,0 +1,27 @@
...
+from errbot import BotPlugin, re_botcmd
+
+
+class The_rules(BotPlugin):
+    """
+    List the bot rules
+    """
+
+    RULES = [
+        "A robot may not harm humanity, or, by inaction, allow humanity to "
```
Udayan12167: Double quotes to single quotes. Quotes bear not throwing errors.

plugins/the_rules.py
```diff
@@ -0,0 +1,27 @@
...
+        "A robot must protect its own existence as long as such protection "
+        "does not conflict with the First or Second Law."
+    ]
+
+    @re_botcmd(pattern=r"the\s+rules", flags=re.IGNORECASE)
+    def the_rules(self, msg, args):
+        """
+        Show the bot rules.
+        """
+        return "\n".join(str(i) + ". " + j for i, j in enumerate(self.RULES))
```
Udayan12167: Is it possible to do it with markdown support? Like are the bullets automatically detected as markdown?

plugins/the_rules.py
meetmangukiya: Quotes bear is set to double quotes though, it'll complain :/

plugins/the_rules.py
meetmangukiya: It is markdown, they will be rendered as order list, not bullets though since this is ordered list

meetmangukiya:
> Can you package both files in a folder inside the plugins folder. Further plugins might have multiple files. It would be better to package that way.

I don't think any plugin will have multiple files. I don't see any plugins extending over 300 lines, but I have no problem with this, I'll make a folder for all the plugins in that case. I'll move the plugins already in master to follow the same convention :+1:

plugins/the_rules.py
meetmangukiya: Not applying here, since applied in other PR, will merge that first and then this, so there are less conflicts :+1:

meetmangukiya: Done :+1:

plugins/the_rules.py
meetmangukiya: > Is it possible to do it with markdown support? Like are the bullets automatically detected as markdown?
  
1. We return valid markdown
2. gitter does the rendering of our markdown

so yes, it is rendered as markdown :+1:

Udayan12167:
rebase needed.

Udayan12167:
ack https://github.com/coala/corobo/pull/52/commits/6dec2cfb8fc63cdbbfd8a88e1406ed04fb7d50ec

meetmangukiya:
@rultor merge!

PR #227, opened 2017-07-25
pitchfork: Add pitchfork plugin

meetmangukiya:
Closes https://github.com/coala/corobo/issues/36

# Reviewers Checklist

### DIFF
--- a/plugins/pitchfork.py
+++ b/plugins/pitchfork.py
@@ -1,0 +1,39 @@
+import re
+import string
+import textwrap
+
+from errbot import BotPlugin, botcmd
+
+
+class Pitchfork(BotPlugin):
+    """
+    To pitchfork users down to ...
+    """
+
+    @botcmd
+    def pitchfork(self, msg, arg):
+        """
+        To pitchfork user down to ...
+        """
+        match = re.match(r'@?([\w-]+)(?:\s+(?:down\s+)?to\s+(.+))?$',
+                         arg)
+        if match:
+            user = match.group(1)
+            place = match.group(2) if match.group(2) else 'offtopic'
+            return textwrap.dedent((
+                string.Template("""
+                    @$user, you are being pitchforked down to $place
+                    ```
+                                                          .+====----->
+                                                           \('
+                    =====================================<%{%{%{>>+===---> $user
+                                                           /(,
+                                                          .+====----->
+                    ```
+                """).substitute(user=user,
+                                place=('[offtopic]('
+                                       'https://gitter.im/coala/coala/offtopic)'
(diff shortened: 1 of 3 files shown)


plugins/pitchfork.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        To pitchfork user down to ...
+        """
+        match = re.match(r'@?([\w-]+)(?:\s+(?:down\s+)?to\s+(.+))?$',
+                         arg)
+        if match:
+            user = match.group(1)
+            place = match.group(2) if match.group(2) else 'offtopic'
+            return (
+                string.Template("""
```
NiklasMM: I find unindented multi line strings disrupt the flow when reading a function/file. You could use [dedent](https://docs.python.org/3.6/library/textwrap.html#textwrap.dedent) to have the literal be properly indented and still not have the indentation in the actual string.

However then you'll probably run into the line length limit.... Hmm.... maybe make the pitchfork a bit shorter? I did notice that on mobile I just saw the stick: 
screenshot_20170726-083940

meetmangukiya:
@NiklasMM PTAL?

plugins/pitchfork.py
```diff
@@ -0,0 +1,36 @@
...
+        To pitchfork user down to ...
+        """
+        match = re.match(r'@?([\w-]+)(?:\s+(?:down\s+)?to\s+(.+))?$',
+                         arg)
+        if match:
+            user = match.group(1)
+            place = match.group(2) if match.group(2) else 'offtopic'
+            return textwrap.dedent((
+                string.Template("""
+                    @$user, you are being pitchforked down to $place
```
SanketDG: if its oftopic, we could make it a markdown link? that links to the offtopic channel

meetmangukiya:
@SanketDG PTAL

plugins/pitchfork.py
```diff
@@ -0,0 +1,39 @@
...
```
SanketDG: any reason to use `string.Template` over https://pyformat.info?

meetmangukiya: yes, because the string contains `{` and `}` and there's no way to escape those(can't believe it)

SanketDG: um you can put `{{` `}}` to escape them, but this is fine, I guess

meetmangukiya:
@SanketDG can you approve if there's nothing else left?

SanketDG:
ack f7363b0

meetmangukiya:
reack 28f72fb

meetmangukiya:
@rultor merge

Issue #233, opened 2017-07-27
Command "corobo" not found if command not found


meetmangukiya:

meetmangukiya:
see https://github.com/errbotio/errbot/issues/1070

PR #235, opened 2017-07-27
config: Suppress not found command

meetmangukiya:
Related to https://github.com/coala/corobo/issues/233

# Reviewers Checklist


config.py
```diff
@@ -45,4 +45,6 @@
 
 CHATROOM_PRESENCE = os.environ.get('ROOMS', '').split() or ROOMS_TO_JOIN
 
+SUPPRESS_CMD_NOT_FOUND = True
```
Udayan12167: Can't we insert a custom reply?

meetmangukiya:
Fixed upstream

PR #268, opened 2017-08-12
Meet/rehelp

meetmangukiya:
# Reviewers Checklist

### DIFF
--- a/plugins/coatils.py
+++ b/plugins/coatils.py
@@ -35,7 +35,8 @@ class Coatils(BotPlugin):
 
         return set(all_langs)
 
-    @re_botcmd(pattern=r'(?:(contrib|bear|lang)\s+)?stats(.+)?(?:(?:\s+)|$)')
+    @re_botcmd(pattern=r'(?:(contrib|bear|lang)\s+)?stats(.+)?(?:(?:\s+)|$)',
+               re_cmd_name_help='(contrib|bear|lang) stats [username|language]')
     def contrib_stats(self, msg, match):
         """
         Allowed commands:
@@ -101,7 +102,8 @@ class Coatils(BotPlugin):
                    ''.format(Coatils.total_bears(),
                              len(Coatils.all_langs())))
 
-    @re_botcmd(pattern=r'ls\s+bears\s+((?:[\w\+]+(?:\s+)?)+)')
+    @re_botcmd(pattern=r'ls\s+bears\s+((?:[\w\+]+(?:\s+)?)+)',
+               re_cmd_name_help='ls bears [langs]+')
     def ls(self, msg, match):
         """
         List bears of given languages:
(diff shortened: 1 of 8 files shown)


plugins/coatils.py
```diff
@@ -163,7 +165,8 @@ def position(stl, stc, enl, enc):
                                                               enl, enc)
 
     # Ignore PycodestyleBear, LineLengthBear
-    @re_botcmd(pattern=r'^run\s+(\w+)((?:\s+\w+(?:\s+\w+=\w+)*)+)\n+```\n([\s\S]+)\n```$')
+    @re_botcmd(pattern=r'^run\s+(\w+)((?:\s+\w+(?:\s+\w+=\w+)*)+)\n+```\n([\s\S]+)\n```$',
+               re_cmd_name_help='run <Bear [[set=ing]+]>+\n```\n<code>+\n```')
```
SanketDG: looks confusing, what's the `=` for?

meetmangukiya: to separate the setting name from the setting value, consider it as a delimiter. Similar to what we have in coafiles

SanketDG: uh okay, why can't we then have `setting=value`

meetmangukiya: we can, I'll change it :+1: 
(Was trying to not blow 80 chars limit, but if it confuses you, it'll confuse others, I'll continue on second line :) )

meetmangukiya:
@SanketDG PTAL?

plugins/coatils.py
```diff
@@ -163,7 +165,9 @@ def position(stl, stc, enl, enc):
                                                               enl, enc)
 
     # Ignore PycodestyleBear, LineLengthBear
-    @re_botcmd(pattern=r'^run\s+(\w+)((?:\s+\w+(?:\s+\w+=\w+)*)+)\n+```\n([\s\S]+)\n```$')
+    @re_botcmd(pattern=r'^run\s+(\w+)((?:\s+\w+(?:\s+\w+=\w+)*)+)\n+```\n([\s\S]+)\n```$',
+               re_cmd_name_help='run <Bear [[setting=value]+]>+\n'
+                                '```\n<code>+\n```')
```
SanketDG: I'd remove the `<>` to avoid any confusion with html tags

Also, even someone experienced using the bot with help text `[setting=value]+]>+` would fail to decipher it

meetmangukiya: hmm.. how else do I convey that "bear name followed by settings" should occur at least once(hence `<>`), and providing settings is optional, hence `[]`

Same for the code, hence `<code>`

SanketDG:
```
run <Bear setting=value....>
code here
```

This is way simpler. Someone using the bear for the first time does not need to know the settings are optional, you could even mention it later in the docs

Another option: Keep as it is but give concrete examples of all possiblities of the regex

meetmangukiya: The docstring of this function contains that, which is also appended to this

Udayan12167:
ack https://github.com/coala/corobo/pull/268/commits/d87a17cc1cc2e6bcfb51508575fd3a78133370d3

meetmangukiya:
reack 809c97fab11507a3f9577fda3a174286ed10ee62 f64b7617c6125c2ceb792a54e28d13a1d7ac9849

meetmangukiya:
@rultor merge

PR #297, opened 2017-10-02
Makman2/more issue formats

Makman2:


Makman2:
Hmm it seems to be more complicated than expected to get the room name...

plugins/labhub.py
```diff
@@ -242,14 +242,63 @@ def mark_cmd(self, msg, match):
...
+            issue_reference_match = re.fullmatch(issue_rgx, issue_reference)
+
+            if issue_reference_match is None:
+                return None
+
+            roomname_rgx = r'(.+?)/(.+)'
+            roomname_match = re.fullmatch(
+                roomname_rgx, msg.frm.room.idd, re.IGNORECASE)
+
+            if roommatch is None:
```
meetmangukiya: You mean `roomname_match`, right?

plugins/labhub.py
```diff
@@ -242,14 +242,63 @@ def mark_cmd(self, msg, match):
...
+                return None
+
+            return roommatch.group(1), roommatch.group(2), m.group(1)
```
meetmangukiya: Same as above

meetmangukiya: There is no `m` in this scope :/

plugins/labhub.py
```diff
@@ -242,14 +242,63 @@ def mark_cmd(self, msg, match):
...
+        def process_super_short_ref(issue_reference):
+            issue_rgx = r'#(\d+)'
+            issue_reference_match = re.fullmatch(issue_rgx, issue_reference)
+
+            if issue_reference_match is None:
+                return None
+
+            roomname_rgx = r'(.+?)/(.+)'
+            roomname_match = re.fullmatch(
+                roomname_rgx, msg.frm.room.idd, re.IGNORECASE)
```
meetmangukiya: `msg.frm.room.uri` instead of `msg.frm.room.idd`

meetmangukiya: However, for uri, you'll find that many coala room URIs are of `coala/coala/<name>` type

plugins/labhub.py
Makman2: ah yeah this has to be the very first match variable from above, will correct it :3
(you already see that my tests aborted above because accessing the room name raised an exception :D)

plugins/labhub.py
Makman2: is that for custom rooms? Or do even repo-related rooms have a name like e.g. `coala/coala/coala-bears`?

Makman2: Also, `room` itself does not exist inside the tests, I'm not sure how to mock it correctly, as this field seems to be backend-dependent.

PR #301, opened 2017-10-16
ship_it: Add more squirrel images

swapagarwal:
# Reviewers Checklist

### DIFF
--- a/plugins/ship_it.py
+++ b/plugins/ship_it.py
@@ -11,10 +11,13 @@ class Ship_it(BotPlugin):
 
     # start ignoring LineLengthBear PyCodeStyleBear
     IMAGES = [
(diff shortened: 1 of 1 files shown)


swapagarwal:
@prnvdixit I thought of that issue as being always open, as we can never have enough of them (squirrels). 😛

prnvdixit:
@swapagarwal Yeah - that 😆....
But still, we need to add ```some more``` images in a patch for now 😃. Plus, we can always add others when even the new ones go 400'd. Still, your call 😄 - You can either recommend these to the ```assignee``` or keep these as they are (I will recommend former though). 😉

swapagarwal:
@prnvdixit Well, I'd prefer suggesting these to the assignee as that is the official way to contribute. 😄

Issue #291, opened 2017-09-13
prevent duplicate issue title


co-robo:
Opened by @jayvdb at [gitter/coala/corobo](https://gitter.im/coala/corobo?at=59b89035bc46472974282df0)

meetmangukiya:
Too heavy on requests I think, especially for coala and coala-bears repo, unless we store the data in db

jayvdb:
Isnt it one search request, and only the top results are needed?  Then post-process the results to look for an identical result.

Anyway, if you need a DB, I suggest that gitmate holds the DB and corobo asks it to do the work 
https://gitlab.com/gitmate/GitMate-2-ideas/issues/17

sils:
shold be one search request, especially now with my latest improvements to IGitt you won't even have to waste a request if you work with the single issues since the partial data from the search response is reused

manu-chroma:
hey, I'd like to work on this one. I am not able to find any documentation for ``IGitt`` apart from the docstrings. Is that the complete documentation or am I looking at the wrong place?

sils:
Yes, that's all the docs we have ATM, sorry

PR #332, opened 2017-11-11
[WIP] labhub.py: Prevent duplicate issue title

manu-chroma:
Before opening a new issue, check for duplicate
issue title in the repo and prevent from opening
if there's any.

Taking a shot at https://github.com/coala/corobo/issues/291. Am I on the right track?

cc: @meetmangukiya @jayvdb 

# Reviewers Checklist

### DIFF
--- a/plugins/labhub.py
+++ b/plugins/labhub.py
@@ -162,6 +162,12 @@ class LabHub(BotPlugin):
 
         if repo_name in self.REPOS:
             repo = self.REPOS[repo_name]
+            # TODO: Return the url to the issue with duplicate issue title
+            # in the return string.
+            for issue in list(repo.search_issues()):
+                if iss_title == issue.title:
+                    return ('Can\'t create an issue with the given title ')
+
             iss = repo.create_issue(iss_title, iss_description + extra_msg)
             return 'Here you go: {}'.format(iss.url)
         else:


yashovardhanagrawal:
I think you're on the right track, but it's not necessary for a person to write the exact same title for a duplicate issue
can we do something for that? 
maybe there's some tool available that can sort this out...

meetmangukiya:
@manu-chroma i don't particularly like comparing the title with ever issue in the issue tracker. This way each time I file a new issue in coala repo, at least 500 issues will be compared :O and also, we'll hit API rate limit way before that.

SanketDG:
This is not a good way of checking for duplicate issues, there are a lot of things like caching and nlp implemented to detect duplicate issues effficiently.

manu-chroma:
Thanks for the feedback everyone. 

@meetmangukiya I'm making one API call per try for opening a new issue, so I don't it should hit the API limit. 
@SanketDG I agree simple checking is not a good approach. Can you link me to any NLP/caching tools which I might use in this situation?

meetmangukiya:
@manu-chroma you are searching issues for complete repo, you are essentially making one API call for each issue data(if the data is not available in the response of search call) and more calls for each page(in pagination). 

I think you might need to inspect `search_issues` function in `IGitt`. I'll take a good look at it later

meetmangukiya:
@manu-chroma do you still want to work on this PR, else close it. :)

PR #340, opened 2017-11-19
filters.py: Add note help diverted to private

### LINKED ISSUE #145: Send long responses as direct messages only with short note in room
Opened via [gitter](https://gitter.im/coala/coala/?at=594a43a1bf1bfe6770d43778) by @jayvdb

yashovardhan:
This leaves a message that 'help' command
is diverted to private message

Closes https://github.com/coala/corobo/issues/145

# Reviewers Checklist

### DIFF
--- a/utils/filters.py
+++ b/utils/filters.py
@@ -27,3 +27,12 @@ class Filters(BotPlugin):
         if msg.frm.nick in self.bot_config.IGNORE_USERNAMES:
             return None, None, None
         return msg, cmd, args
+
+    @cmdfilter
+    def filter_private_diverts(self, msg, cmd, args, dry_run):
+        """
+        Leave a message that response is diverted as private message.
+        """
+        if cmd in self.bot_config.DIVERT_TO_PRIVATE:
+            self.send(msg.frm, '@{}, help message is diverted into private, check DM with @co-robo')
+        return msg, cmd, args


utils/filters.py
```diff
@@ -27,3 +27,12 @@ def filter_ignored_users(self, msg, cmd, args, dry_run):
...
         return msg, cmd, args
+
+    @cmdfilter
+    def filter_private_diverts(self, msg, cmd, args, dry_run):
+        """
+        Leave a message that response is diverted as private message.
+        """
+        if cmd in self.bot_config.DIVERT_TO_PRIVATE:
+            self.send(msg.frm, 'It is diverted into private message, try sending this command directly to @co-robo')
+        return msg, cmd, args
```
yashovardhanagrawal: this code is working fine but it is blocked due to an issue in errbot

[errbotio/errbot#1071](https://github.com/errbotio/errbot/issues/1071)

utils/filters.py
```diff
@@ -27,3 +27,12 @@ def filter_ignored_users(self, msg, cmd, args, dry_run):
...
             return None, None, None
         return msg, cmd, args
+
+    @cmdfilter
+    def filter_private_diverts(self, msg, cmd, args, dry_run):
+        """
+        Leave a message that response is diverted as private message.
+        """
+        if cmd in self.bot_config.DIVERT_TO_PRIVATE:
+            self.send(msg.frm, 'It is diverted into private message, check DM with @co-robo')
```
jayvdb: per https://github.com/coala/corobo/issues/349 , I think this message should refer to the caller, so they know it is a response to them.  (otherwise everyone goes looking for a DM.)

jayvdb (changes requested):

jayvdb:
Do you know why I added `blocked` label?? I've forgotten ;-)

utils/filters.py
yashovardhanagrawal: okay.. I will fix that

yashovardhanagrawal:
Done with the changes

jayvdb:
sort of blocked on https://github.com/errbotio/errbot/pull/1139

Actually the better local solution is to cut the size of the help down to one gitter message size.  We have bugs about that.

We could also have a CI rule which verifies the length of the help (for normal users) is shorter than the gitter message length.

jayvdb:
unack https://github.com/coala/corobo/pull/340/commits/22dae6193ff7ae1ef53a71232a0885a9dc6b3a2f

jayvdb:
errbotio/errbot#1139 was merged

meetmangukiya:
> errbotio/errbot#1139 was merged

Actually this still cannot be fixed. partial property is set for all messages that were broken down from a single message. There's no way to determine if these messages belong to a particular group, or if the partial message is the first message. Either of which would be required to stop the second divert message.

Issue #354, opened 2017-11-22
Exception using Text backend with default config


co-robo:
```
  File "/home/jayvdb/projects/coala/dependencies/errbot/errbot/backends/text.py", line 214, in __init__
    self.user = self.build_identifier(self.bot_config.BOT_ADMINS[0])
  File "/home/jayvdb/projects/coala/dependencies/errbot/errbot/backends/text.py", line 374, in build_identifier
    raise ValueError('An identifier for the Text backend needs to start with # for a room or @ for a person.')
```

The default `BOT_ADMINS` is `['meetmangukiya']`
The default `BOT_ADMINS` first entry needs to start with `@`, as starting with `#` provokes https://github.com/errbotio/errbot/issues/1135
difficulty/low
Opened by @jayvdb at [Gitter](https://gitter.im/coala/corobo?at=5a1575b7982ea2653faf889b)

Issue #353, opened 2017-11-22
Exception at start if default data directory does not exist


co-robo:
And the defaul data directory does not exist
Opened by @jayvdb at [Gitter](https://gitter.im/coala/corobo?at=5a156dc4232e79134dbabd18)

PR #360, opened 2017-11-22
Make the bot usable on startup without configuration

### LINKED ISSUE #352: Exception at startup if gitter token not provided
Opened by @jayvdb at [Gitter](https://gitter.im/coala/corobo?at=5a156c85982ea2653faf51dd)

jayvdb:
Fixes #352, #353 & #354

### DIFF
--- a/config.py
+++ b/config.py
@@ -24,7 +24,7 @@ if not os.environ.get('COBOT_PREFIX'):
 BOT_ADMINS = os.environ.get('BOT_ADMINS', '').split() or ('meetmangukiya', )
 
 BOT_IDENTITY = {
-    'token': os.environ['COBOT_TOKEN']
+    'token': os.environ.get('COBOT_TOKEN')
 }
 
 IGNORE_USERNAMES = os.environ.get("IGNORE_USERNAMES",
--- a/config.py
+++ b/config.py
@@ -3,7 +3,7 @@ import os
 
 COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
 
-BACKEND = os.environ.get('BACKEND', 'Gitter')
+BACKEND = os.environ.get('BACKEND', 'Text')
 
 BOT_EXTRA_BACKEND_DIR = os.path.join(COBOT_ROOT, 'err-backend-gitter')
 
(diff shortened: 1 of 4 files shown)


config.py
```diff
@@ -21,7 +21,7 @@
 if not os.environ.get('COBOT_PREFIX'):
     BOT_ALT_PREFIXES = ('cobot ', )
 
-BOT_ADMINS = os.environ.get('BOT_ADMINS', '').split() or ('meetmangukiya', )
+BOT_ADMINS = os.environ.get('BOT_ADMINS', '').split() or ('@corobo', )
```
meetmangukiya: current `corobo` is actually @co-robo and the bot account shouldn't be its own admin IMO.

jayvdb:
You're kidding me?  You are using this as the default for the production server?

Is there a way I can detect the production server (hostname?) in order to not break it?

config.py
jayvdb: This value has nothing to do with the account `@co-robo` used by a production instance.

This is what other people 's test instance has as a value for this field.

I am a bit interested in whether the `@` there causes any problems.

yukiisbored:
@jayvdb No, there's no way to check if it's broken or not other than testing corobo. Due to the amount of RAM we got, we can't startup a whole grafana setup

config.py
jayvdb: The `@` there does no harm, but according to the docs, `corobo@localhost` is going to be safer

config.py
```diff
@@ -21,10 +21,10 @@
 if not os.environ.get('COBOT_PREFIX'):
     BOT_ALT_PREFIXES = ('cobot ', )
 
-BOT_ADMINS = os.environ.get('BOT_ADMINS', '').split() or ('meetmangukiya', )
+BOT_ADMINS = os.environ.get('BOT_ADMINS', '').split() or ('*@localhost', )
```
meetmangukiya: According to errbot specs, for text backend, it should be `@localhost` or `#localhost`

meetmangukiya:
ack c79e25dbe90300dc168c30ece7e8d103ef3361a0 b8c691fae374311bd391d17e15104fe6b2057d11

config.py
jayvdb: No it isnt.  See `config-template.py`

jayvdb: See also https://github.com/errbotio/errbot/issues/1135

meetmangukiya: okk :+1:

meetmangukiya:
ack da63e1024ccda0e7df466630240b150ad5f63f0f

meetmangukiya:
@rultor merge

config.py
jayvdb: Sorry, you were right for the `Text` backend.  We should fix that here quickly, and upstream, as it is silly that `Text` doesnt accept the normal syntax when it shouldnt care about the bot admin anyway as it isnt on a shared service.

Issue #373, opened 2017-11-24
Rewrite deprecate_cobot to be generic deprecate_bot_prefix


jayvdb:
The module and function names shouldnt mention `cobot` or `corobo`.
It should default to the values in `BOT_DEPRECATED_PREFIXES`

And then the entry here can be updated to be more appealing to other errbot users:

https://github.com/errbotio/errbot/wiki#16---deprecate_cobot

PR #400, opened 2017-11-27
plugins: Rewrite deprecate_cobot to be generic

mamat-rahmat:
This make plugin generic to any deprecated prefix.

Closes https://github.com/coala/corobo/issues/373

# Reviewers Checklist


utils/deprecate_bot_prefixes.py
```diff
@@ -0,0 +1,21 @@
...
+    A callback for every message that starts with depecrated prefix, hence, leaving a deprecation notice
+    """
+
+    def callback_message(self, msg):
+        """
+        Notify the user issuing the command that use deprecated prefix.
+        """
+
+        for deprecated_prefix in config.BOT_DEPRECATED_PREFIXES:
+            if msg.body.startswith(deprecated_prefix+" "):
```
jayvdb: I think `+" "` isnt needed here, as the prefix will already contain a ` `

config.py
```diff
@@ -21,6 +21,9 @@
 if not os.environ.get('COBOT_PREFIX'):
     BOT_ALT_PREFIXES = ('cobot ', )
 
+BOT_DEPRECATED_PREFIXES = os.environ.get(
+    'BOT_DEPRECATED_PREFIXES', '').split() or ('cobot', )
```
jayvdb: The prefix should be `cobot ` .  The default is `!` , and it is possible that people might want to use this plugin to deprecate `!`, without any spaces being added after it.

jayvdb:
Also, could you move this into `plugins/` , as it is now a normal re-usable plugin.

utils/deprecate_bot_prefixes.py
mamat-rahmat: I got inspiration from how config define alternative for BOT_PREFIX. It use `'corobo '` instead of `'corobo'`. It respond to `'corobo command'` and not `'corobocommand'`. How about that?

Because i use whitespace for splitting env var, then i hardcoded space to separate prefix and command.

mamat-rahmat: Ah, i understand from above case. Thanks!

config.py
mamat-rahmat: I dont understand this

> The prefix should be cobot.

Do you mean we should hardcode it instead of getting from env var?

jayvdb: Oh, I am sorry ... github isnt rendering that correctly.  `'cobot' + ' '`

plugins/deprecate_bot_prefixes.py
```diff
@@ -0,0 +1,21 @@
+from errbot import BotPlugin
+import config
```
meetmangukiya: Remove this import statement

plugins/deprecate_bot_prefixes.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    A callback for every message that starts with depecrated prefix, hence, leaving a deprecation notice
+    """
+
+    def callback_message(self, msg):
+        """
+        Notify the user issuing the command that use deprecated prefix.
+        """
+
+        for deprecated_prefix in config.BOT_DEPRECATED_PREFIXES:
```
meetmangukiya: use `self.bot_config.BOT_DEPRECATED_PREFIXES` instead of `config.BOT_DEPRECATED_PREFIXES`

plugins/deprecate_bot_prefixes.py
```diff
@@ -0,0 +1,21 @@
...
+            if msg.body.startswith(deprecated_prefix):
+                self.send(
+                    msg.frm,
+                    "@{} usage of {} has been deprecated, please use {} "
+                    "from now on.".format(msg.frm.nick, deprecated_prefix,
+                                          config.BOT_PREFIX)
```
meetmangukiya: Change accordingly, as described above

mamat-rahmat:
@meetmangukiya Changes applied. Tests now passed. But, i'm not really sure did it right.

jayvdb:
Change `Fixes ` to `Closes ` .  This isnt a bug; it is an enhancement .

mamat-rahmat:
Done

tests/deprecate_bot_prefixes_test.py
```diff
@@ -0,0 +1,14 @@
+import os
+import config
+
+pytest_plugins = ['errbot.backends.test']
+extra_plugin_dir = 'plugins'
+
+def test_deprecated_prefixes_default(testbot):
```
meetmangukiya: Remove this test case, not needed. The other test case is enough

tests/deprecate_bot_prefixes_test.py
```diff
@@ -0,0 +1,14 @@
+import os
+import config
```
meetmangukiya: Remove this import, it won't allow to run tests from anywhere other than root 😕

mamat-rahmat:
Done

jayvdb:
ack https://github.com/coala/corobo/pull/400/commits/2c2ae5e0685d5629820a0c10101395dc03bb7357

jayvdb:
@rultor merge

PR #403, opened 2017-11-27
Re_botcmd_fullmatch: Add re_botcmd_fullmatch

meetmangukiya:
Closes https://github.com/coala/corobo/issues/402

# Reviewers Checklist


supergr35:
test coverage ;)

plugins/re_botcmd_fullmatch.py
```diff
@@ -0,0 +1,35 @@
+import time
+import threading
+
+from errbot import BotPlugin, botcmd
+
+
+class Re_botcmd_fullmatch(BotPlugin):
```
jayvdb: rather than a plugin, we should not use `re_botcmd` , and instead use a better decorator which does this check and then defers to `re_botcmd`

Issue #333, opened 2017-11-13
Ping PR author when marking WIP


co-robo:
difficulty/low
Opened by @jayvdb at [Gitter](https://gitter.im/coala/coala?at=5a0989d919147ac323197687)

jayvdb:
To work on any backend, this requires fetching the identity of the PR author, hence moving to medium difficulty.

PR #415, opened 2017-11-29
labhub.py: Ping PR author when marking WIP

andrewda:
This will ping the author of a PR when it is marked as WIP through the
`corobo mark wip <issue>` command.

Closes https://github.com/coala/corobo/issues/333

# Reviewers Checklist

### DIFF
--- a/plugins/labhub.py
+++ b/plugins/labhub.py
@@ -214,6 +214,12 @@ class LabHub(BotPlugin):
             return 'Repository doesn\'t exist.'
         else:
             current_labels = list(mr.labels)
+            data = mr.data
+            login = None
+            if 'user' in data:
+                login = data['user']['login']
+            elif 'author' in data:
+                login = data['author']['username']
             if state == 'wip':
                 pending_labels = ['process/pending_review',
                                   'process/pending review']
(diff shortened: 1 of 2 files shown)


plugins/labhub.py
```diff
@@ -214,6 +214,8 @@ def mark_cmd(self, msg, match):
             return 'Repository doesn\'t exist.'
         else:
             current_labels = list(mr.labels)
+            data = mr._get_data()
```
meetmangukiya: Please don't use private functions. Since, it is not mandatory that you'll get a response of same schema from all providers like GitHub, GitLab, etc.

andrewda:
@meetmangukiya I opened https://gitlab.com/gitmate/open-source/IGitt/merge_requests/159 upstream to add the `author` property. When that gets merged I'll fix it here!

jayvdb:
Could you try to build a test for this, especially a test which uses a GitLab URL ;-)

plugins/labhub.py
```diff
@@ -225,9 +231,11 @@ def mark_cmd(self, msg, match):
                 return ('The pull request {mr_link} is marked *work in progress'
                         '*. Use `{bot_prefix} mark pending` or push to your '
                         'branch if feedback from the community is needed '
-                        'again.'.format(
+                        'again.\n@{user_login}, please check your pull '
+                        'request.'.format(
                             mr_link=mr.url,
-                            bot_prefix=self.bot_config.BOT_PREFIX)
+                            bot_prefix=self.bot_config.BOT_PREFIX,
+                            user_login=login)
```
jayvdb: if `login` is `''` , the markdown will be `@,` , which would be a bug.  So a small fix is needed here.

Otherwise this looks great.  Could you squash it into a single commit when you do your next push.

andrewda: What would you suggest doing here? Should I just omit the ping if the login can't be found?

jayvdb: That is probably the best approach.

jayvdb: And so you dont think I am crazy ..
There are lots of reasons why the username may not exist.  I suspect that this isnt very likely to occur when an MR is first created, but we have people vanish after they create their MR.  GitHub now has a @ghost account which it substitutes in, but I dont know what GitLab and other services respond to these odd scenarios (corobo will also support BitBucket and other services soon also).

jayvdb:
ack https://github.com/coala/corobo/pull/415/commits/f2d506e4451ac7c6d4daf55675dd8fef4fe2d990

jayvdb:
@rultor merge

Issue #171, opened 2017-06-24
List rebased PRs


co-robo:
Ported from old cobot repo, opened by @jayvdb - https://gitlab.com/coala/cobot/issues/31
labels: plugin proposal, importance/medium, difficulty/high, cool and awesome

jayvdb:
There is some important discussion that occurred at https://gitlab.com/coala/cobot/issues/31

Mixih:
Assigned per GCI task

PR #423, opened 2017-11-30
git_stats.py: New plugin

LifetimeAlone:
labhub.py: list rebased PRs

Add list mergre-ready PRs command
'pr list \<complete-repo-URL\>'.
Lists open PRs, which don't need
rebase and have 'process/approved'
or 'process/review_pending' labels.

Closes https://github.com/coala/corobo/issues/171

# Reviewers Checklist

### DIFF
--- a/plugins/git_stats.plug
+++ b/plugins/git_stats.plug
@@ -1,0 +1,11 @@
+[Core]
+name = git_stats
+module = git_stats
+
+[Documentation]
+description = GitHub and GitLub statistics
+
+[Python]
+version = 3
+
+[Errbot]
--- a/plugins/git_stats.py
+++ b/plugins/git_stats.py
@@ -1,0 +1,78 @@
+import datetime
+import re
+from shutil import rmtree
+
+from errbot import re_botcmd
+
+from plugins.labhub import LabHub
+
+
+class GitStats(LabHub):
+    """GitHub and GitLab statistics"""  # Ignore QuotesBear
+
+    PR_LABELS = ('process/approved',
+                 'process/pending review',
+                 'process/pending-review'
+                 )
+
+    def __init__(self, bot, name=None):
+        super().__init__(bot, name)
+
+    @re_botcmd(pattern=r'mergable\s+([^/]+)',  # Ignore PyCodeStyleBear
+               re_cmd_name_help='pr list <repo>',
+               flags=re.IGNORECASE)
+    def pr_list(self, msg, match):
+        """List PRs ready to be merged."""  # Ignore QuotesBear
+        repo_name = match.groups(1)[0]
+
+        try:
+            merge_requests = self.REPOS[repo_name].merge_requests()
+        except KeyError:
+            return "Repository doesn't exist."
+
+        checks = []
+
+        def check_mr(func):
+            checks.append(func)
+            return func
+
+        @check_mr
+        def check_labels(merge_request):
+            labels = set(merge_request.labels)
+            if labels.intersection(self.PR_LABELS):
+                return True
+            return False
+
+        @check_mr
+        def check_state(merge_request):
+            if merge_request.state == 'open':
+                return True
+            return False
+
+        @check_mr
+        def check_rebased(merge_request):
+            repo, tempdir = merge_request.repository.get_clone()
+            head_sha = repo.head.commit.hexsha
+            rmtree(tempdir)
+            if merge_request.base.sha == head_sha:
+                return True
+            return False
+
+        def merge_ready(merge_request):
+            for chk in checks:
+                if not chk(merge_request):
+                    return False
+            return True
+
+        ready_to_merge = []
+        for mr in merge_requests:
+            if merge_ready(mr):
+                ready_to_merge.append(mr)
+        if len(ready_to_merge) == 0:
+            return "No merge-ready PRs!"
+        else:
+            now = datetime.datetime.now()
+            response = "PRs ready to be merged:\n\n"
+            for mr in sorted(ready_to_merge, key=lambda x: now - x.created):
+                response += '{}\n'.format(mr.url)
+            return response
--- a/tests/git_stats_test.py
+++ b/tests/git_stats_test.py
@@ -1,0 +1,103 @@
+import logging
+from tempfile import mkdtemp
+import unittest
+from unittest.mock import Mock, MagicMock, create_autospec, PropertyMock, patch
+
+from IGitt.GitHub.GitHubMergeRequest import GitHubMergeRequest
+from IGitt.GitLab.GitLabMergeRequest import GitLabMergeRequest
+from IGitt.GitHub.GitHubIssue import GitHubIssue
+from IGitt.GitLab.GitLabIssue import GitLabIssue
+from git import Repo
+
+import github3
+import IGitt
+import plugins.git_stats
+from tests.helper import plugin_testbot
+
+
+class TestGitStats(unittest.TestCase):
+
+    def setUp(self):
+        plugins.git_stats.github3 = create_autospec(github3)
+        self.mock_org = create_autospec(github3.orgs.Organization)
+        self.mock_gh = create_autospec(github3.GitHub)
+        self.mock_repo = create_autospec(IGitt.GitHub.GitHub.GitHubRepository)
+        plugins.git_stats.github3.login.return_value = self.mock_gh
+        self.mock_gh.organization.return_value = self.mock_org
+        plugins.git_stats.github3.organization.return_value = self.mock_org
+
+    def test_pr_list(self):
+        git_stats, testbot = plugin_testbot(plugins.git_stats.GitStats, logging.ERROR)
+        git_stats.activate()
+
+        git_stats.REPOS = {'test': self.mock_repo}
+        mock_github_mr = create_autospec(GitHubMergeRequest)
+        mock_gitlab_mr = create_autospec(GitLabMergeRequest)
+        mock_github_issue = create_autospec(GitHubIssue)
+        mock_gitlab_issue = create_autospec(GitLabIssue)
+        mock_github_mr.closes_issue = mock_github_issue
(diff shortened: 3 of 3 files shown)


plugins/labhub.py
```diff
@@ -387,3 +388,63 @@ def pr_stats(self, msg, match):
...
+            checks = []
+
+            def check_mr(function):
+                checks.append(function)
+                return function
+
+            @check_mr
+            def check_labels(mr):
+                labels = mr.closes_issue.labels
+                if 'process/approved' in labels and 'process/review_pending':
```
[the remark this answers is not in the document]
meetmangukiya: Though I think we can rename them to match.

plugins/labhub.py
```diff
@@ -387,3 +388,63 @@ def pr_stats(self, msg, match):
...
+        else:
+            checks = []
+
+            def check_mr(function):
+                checks.append(function)
+                return function
+
+            @check_mr
+            def check_labels(mr):
+                labels = mr.closes_issue.labels
```
meetmangukiya: Should be checking MR labels `mr.labels` instead

LifetimeAlone: Fixed it!

plugins/labhub.py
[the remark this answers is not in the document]
LifetimeAlone: Added both variants so it will definitely work

LifetimeAlone:
Whoops, will see what's wrong

plugins/labhub.py
LifetimeAlone: There's no mr.labels func in IGitt :(

LifetimeAlone: Maybe I should add an issue in IGitt repo?

meetmangukiya: Are you sure, it should be! `MergeRequest` classes inherit from `Issue` classes, and hence, `labels` property should be available in `MergeRequest`s as well

LifetimeAlone: For some reason, it doesn't work with mr.labels, can't find out why

LifetimeAlone: Seems like I found the bug, will push my changes in a minute :)

LifetimeAlone: Yup, got everything fixed, it works with mr.labels now

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
                     state=type(self).community_state(pr_count)
                  )
         yield reply
+
+    @re_botcmd(pattern=r'pr\s+list\s+https://(github|gitlab)\.com/([^/]+)/([^/]+)',  # Ignore LineLengthBear, PyCodeStyleBear
```
meetmangukiya: i think we can just use repo name. We are discarding the `host` and `org` either way. And this way we'd only need to give repo name and not the complete url. 

Just like `corobo new issue <repo> ..`

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
                     state=type(self).community_state(pr_count)
                  )
         yield reply
+
+    @re_botcmd(pattern=r'pr\s+list\s+(\w+)',  # Ignore LineLengthBear, PyCodeStyleBear
```
meetmangukiya: This wouldn't catch repo names like `coala-bears`, refer to earlier regex

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
...
+               flags=re.IGNORECASE)
+    def pr_list(self, msg, match):
+        """List PRs ready to be merged."""  # Ignore QuotesBear
+        repo_name = match.groups(1)[0]
+
+        try:
+            mrs = self.REPOS[repo_name].merge_requests()
+        except KeyError:
+            return 'Repository doesn\'t exist.'
+        else:
```
jayvdb: this `else:` causes an unnecessary indent. (yes, I know you will have seen it elsewhere ... but that is wrong)

there is a return in the `except:` block above.

And any other exception will cascade up.

And `try: else:` is only useful when one of the exception paths does not exit the function. (and some strange cases including use of `finally:`.  See https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
...
+               re_cmd_name_help='pr list <repo>',
+               flags=re.IGNORECASE)
+    def pr_list(self, msg, match):
+        """List PRs ready to be merged."""  # Ignore QuotesBear
+        repo_name = match.groups(1)[0]
+
+        try:
+            mrs = self.REPOS[repo_name].merge_requests()
+        except KeyError:
+            return 'Repository doesn\'t exist.'
```
jayvdb: use double quotes to contain a single quote.
(or is our QuoteBear causing problems again?)

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
...
+            checks = []
+
+            def check_mr(function):
+                checks.append(function)
+                return function
+
+            @check_mr
+            def check_labels(mr):
+                labels = mr.labels
+                rl = ['process/approved', 'process/pending review',
```
jayvdb: these belong in config somewhere ; at least put them in a CONSTANT at the top of this module.

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
...
+                      'process/pending-review']  # Possible labels
+                if (rl[0] in labels) or (rl[1] in labels) or (rl[2] in labels):
```
jayvdb: use a set for the current PR labels and a set for the configured set of mergable-PR-labels , and then use set intersection.

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
...
+                return False
+
+            def merge_ready(mr):
+                for chk in checks:
+                    if not chk(mr):
+                        return False
+                return True
+
+            rdy = []  # List of ready-to-merge PRs
+            response = 'PRs ready to be merged:\n\n'
```
jayvdb: you do not need `response` here; move it down to where you need it.

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
...
+            def check_labels(mr):
+                labels = mr.labels
+                rl = ['process/approved', 'process/pending review',
+                      'process/pending-review']  # Possible labels
+                if (rl[0] in labels) or (rl[1] in labels) or (rl[2] in labels):
+                    return True
+                return False
+
+            @check_mr
+            def check_rebased(mr):
```
jayvdb: this is an expensive check; it should be the last check in the list of checks.

plugins/labhub.py
```diff
@@ -387,3 +388,65 @@ def pr_stats(self, msg, match):
                     state=type(self).community_state(pr_count)
                  )
         yield reply
+
+    @re_botcmd(pattern=r'pr\s+list\s+([^/]+)',  # Ignore LineLengthBear, PyCodeStyleBear
```
jayvdb: I dont like `corobo pr list` , because this is about mergable PRs only, so the command needs to reflect that.

tests/labhub_test.py
```diff
@@ -324,3 +327,77 @@ def test_invite_me(self):
...
+        labhub.REPOS = {'test': self.mock_repo}
+        mock_github_mr = create_autospec(GitHubMergeRequest)
+        mock_gitlab_mr = create_autospec(GitLabMergeRequest)
+        mock_github_issue = create_autospec(GitHubIssue)
+        mock_gitlab_issue = create_autospec(GitLabIssue)
+        mock_github_mr.closes_issue = mock_github_issue
+        mock_gitlab_mr.closes_issue = mock_gitlab_issue
+        mock_github_mr.repository = self.mock_repo
+        mock_gitlab_mr.repository = self.mock_repo
+        mock_github_mr.url = '1'
```
jayvdb: use a real-ish URL, so the tests are more readable.

plugins/labhub.py
LifetimeAlone: Is `corobo mergable prs` ok?

plugins/labhub.py
LifetimeAlone: Nope, everything is alright, fixed it

plugins/labhub.py
LifetimeAlone: Ok, got you, I see the difference now

plugins/git_stats.py
```diff
@@ -0,0 +1,78 @@
...
+                return True
+            return False
+
+        def merge_ready(mr):
+            for chk in checks:
+                if not chk(mr):
+                    return False
+            return True
+
+        rdy = []  # List of mergable PRs
```
jayvdb: Please use real words as variable names, with 'snake case'.  See PEP 8 coding standard.

LifetimeAlone: Yup, got it fixed

plugins/git_stats.plug
```diff
@@ -0,0 +1,11 @@
...
+name = git_stats
+module = git_stats
+
+[Documentation]
+description = GitHub and GitLub statistics
+
+[Python]
+version = 3
+
+[Errbot]
```
yukiisbored: Missing UNIX new line on EOF

LifetimeAlone: Fixed!

yukiisbored:
ack https://github.com/coala/corobo/pull/423/commits/07fb758f046a15def302f4f3b00408e352bc620b

yukiisbored:
Gonna leave merging to @meetmangukiya / @jayvdb

jayvdb:
@rultor merge

yukiisbored:
silly rultor

yukiisbored:
@rultor Merge this now

Issue #337, opened 2017-11-17
Ask people to not use 'sir' when they do


co-robo:
Opened by @adtac at [Gitter](https://gitter.im/coala/coala?at=5a0ee9fbcc1d527f6b813e40)

RaiVaibhav:
@shaipranesh11 did you first claimed the task ?https://codein.withgoogle.com/dashboard/tasks/6536913504698368/

shaipranesh11:
i claimed it.

PR #431, opened 2017-12-01
creating chat_err_monitor.py to prevent people from using 'sir' and '…

shaipranesh11:
…C' in 'Corobo'

Ask people to not use 'sir' when they do :D.#337. The file was renamed.

# Reviewers Checklist

### DIFF
--- a/plugins/chat_error_monitor.py
+++ b/plugins/chat_error_monitor.py
@@ -9,10 +9,20 @@ class Coala_lowercase_c(BotPlugin):
 
     def callback_message(self, msg):
         emots = [':(', ':angry:', ':confounded:',
-                 ':disappointed:', ':triumph:']
+                 ':disappointed:', ':triumph:' , ':D']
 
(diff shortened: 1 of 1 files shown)


plugins/chat_error_monitor.py
```diff
@@ -9,10 +9,20 @@ class Coala_lowercase_c(BotPlugin):
 
     def callback_message(self, msg):
         emots = [':(', ':angry:', ':confounded:',
-                 ':disappointed:', ':triumph:']
+                 ':disappointed:', ':triumph:' , ':D']
```
andrewda: I don't think a smiley face emoji is a good option here. This is a list of emojis that will be randomly used down on line 30, and a smiley doesn't fit the pattern.

RaiVaibhav: In `match_coala` bot is printing random emoji from a list  of emots, changing emots is not a good idea ,there is also a 20% chance that  `:D`  will be shown if some one write `coala` with uppercase, 
The

PR #447, opened 2017-12-02
creating no_sir.py: to tell users to not use 'sir'

shaipranesh11:
Ask people to not use 'sir' when they do #337.

# Reviewers Checklist

### DIFF
--- a/plugins/no_sir.py
+++ b/plugins/no_sir.py
@@ -1,0 +1,15 @@
+from errbot import BotPlugin
+
+
+class no_sir(BotPlugin):
+    """do not use sir to sound more friendly"""  # Ignore QuotesBear
+
+    def callback_message(self, msg):
+        emots = [':D']
+        if match_sir:
+            self.send(
+                msg.frm,
+                '@{}, donot use \'sir\' in ur conversation. {}'.format(
+                    msg.frm.nick, emots[0]
+                )
+            )


plugins/no_sir.py
```diff
@@ -0,0 +1,15 @@
+from errbot import BotPlugin
+
+
+class no_sir(BotPlugin):
+    """do not use sir to sound more friendly"""  # Ignore QuotesBear
+
+    def callback_message(self, msg):
```
LifetimeAlone: It's better to use `@re_botcmd`, check other modules for some useful patterns :wink:

plugins/no_sir.py
```diff
@@ -0,0 +1,15 @@
...
+        emots = [':D']
+        if match_sir:
+            self.send(
+                msg.frm,
+                '@{}, donot use \'sir\' in ur conversation. {}'.format(
```
LifetimeAlone: This indentation looks quite weird and it's hard to read it, try to make it a bit more compact :)

plugins/no_sir.py
```diff
@@ -0,0 +1,15 @@
+from errbot import BotPlugin
```
LifetimeAlone: Please, add tests and no_sir.plug. 
See other .plug files in plugins/ directory and tests in tests/ directory.
New modules can't be added without .plug and testing :cry:

shaipranesh11:
what's wrong with this?

plugins/no_sir.py
shaipranesh11: k

plugins/no_sir.py
shaipranesh11: k thanks

PR #448, opened 2017-12-02
creating no_sir.py: to tell users to not use 'sir'

shaipranesh11:
Ask people to not use 'sir' when they do #337.

# Reviewers Checklist


jayvdb:
@shaipranesh11 , you must keep working on your branch `patch-9` , and on PR 448, until you have a solution for this problem in *this* PR.

Issue #257, opened 2017-08-09
explain: Move hard coded explanations out of code


co-robo:
It would be nice to be able to add and remove explanations to make it fit for another org.
Opened by @yukiisbored at [gitter//yukiisbored](https://gitter.im//yukiisbored?at=598afb9dbc4647297468dc48)

jayvdb:
This can be done by converting the explanations to use errbot templates.

jayvdb:
The solution for this issue should also use coala to validate the template syntax.

Issue #329, opened 2017-11-09
Fix spelling of "eligility" in labhub.py


prnvdixit:
Change all such occurrences to "eligibility".
label: initiatives/gci

jayvdb:
Closed as part of https://github.com/coala/corobo/issues/257

PR #471, opened 2017-12-05
plugins: Move hard coded messages out of code

andrewda:
This begins to move the large chunks of hard coded messages out of
code and into markdown template files. This will make it easier for other
organizations to adopt corobo and customize it to their needs.

Closes https://github.com/coala/corobo/issues/257

# Reviewers Checklist


andrewda:
AFAIK the markdown templates can't be limited to 80 characters because Gitter renders those extra newlines literally.

plugins/explain.py
```diff
@@ -1,103 +1,40 @@
...
-                self.MSGS.get(
-                    match.group(1).lower(),
-                    self.ERROR_MSG
-                ).format(bot_prefix=self.bot_config.BOT_PREFIX))
+        response = ''
+        filename = 'explanations/{}.jinja2.md'.format(match.group(1).lower())
+        if match.group(1).lower() in self.KNOWN_KEYS:
+            if match.group(2):
+                response += '@{}: \n'.format(match.group(2))
+            response += tenv().get_template(filename).render()
```
jayvdb: you can pass in `username` here, and it can be None, and use jinja's `if` syntax.

Bugger it, pass in `username` to them all, as we have another issue about personalising these messages, so if the username is available, that task becomes a jinja-markdown-only task, so newcomer level.

jayvdb: Hmm, actually, we should have more clarity here, we need the "actor" username, and the "target" username.  They should have a consistent variable name through all of the templates.

plugins/explain.py
```diff
@@ -1,103 +1,43 @@
...
-                            if match.group(2) else '') +
-                self.MSGS.get(
-                    match.group(1).lower(),
-                    self.ERROR_MSG
-                ).format(bot_prefix=self.bot_config.BOT_PREFIX))
+        user = msg.frm.nick
+        response = ''
+        filename = 'explanations/{}.jinja2.md'.format(match.group(1).lower())
+        if match.group(1).lower() in self.KNOWN_KEYS:
+            if match.group(2):
```
jayvdb: This is the username of the person who is the message target.  It should also be a variable, passed into the template

andrewda: Now passed into the template 👍

plugins/templates/explanations/closes.jinja2.md
```diff
@@ -0,0 +1 @@
+We use bug prediction in coala which relies on the `Fixes` keyword in commit messages. To get good results from that we need to use `Closes` for normal issues instead of `Fixes` which should only be used for real bugs. (See also [the commit message docs](https://coala.io/commit).) To change your message you just use `git commit --amend` and then `git push --force` the new commit to replace the old one.
```
meetmangukiya: Too big for one line, please follow 80 cols per line here as well IMO

meetmangukiya: Same for every template file

meetmangukiya: Add a `LineLengthBear` to `jinja` section of coafile

plugins/explain.py
meetmangukiya: Is this intended, none of the two variables are used in any of the templates? Why not just pass it when we need it in the future?

plugins/explain.py
```diff
@@ -1,103 +1,44 @@
...
                re_cmd_name_help='explain <term>',
                flags=re.IGNORECASE)
     def explain(self, msg, match):
         """Explain various terms."""  # Ignore QuotesBear
-        return ('{}'.format('@{}: \n'.format(match.group(2))
-                            if match.group(2) else '') +
-                self.MSGS.get(
-                    match.group(1).lower(),
-                    self.ERROR_MSG
-                ).format(bot_prefix=self.bot_config.BOT_PREFIX))
```
meetmangukiya: @andrewda missing this in the final change, we want to replace `bot_prefix` in one of the strings

plugins/labhub.py
```diff
@@ -129,10 +110,18 @@ def invite_cmd(self, msg, match):
...
+            ).render(
+                username=invitee,
+            )
         else:
-            return ('@{}, you are not a maintainer, only maintainers can invite'
-                    ' other people. Nice try :poop:'.format(inviter))
+            return tenv().get_template(
+                'errors/not-maintainer.jinja2.md'
+            ).render(
+                action='invite other people',
```
meetmangukiya: huh, this is good 👍

plugins/labhub.py
```diff
@@ -333,13 +320,19 @@ def eligible(user, iss):
...
             elif user in iss.assignees:
                 yield ('The issue is already assigned to you.')
             else:
-                yield ('The issue is already assigned to someone. Please '
-                       'check if the assignee is still working on the issue, '
-                       'if not, you should ask for reassignment.')
+                yield tenv().get_template(
+                    'errors/already-assigned.jinja2.md'
+                ).render(
+                    username=user
```
meetmangukiya: since you are following this everywhere else, append a `,` ;)

plugins/templates/explanations/closes.jinja2.md
andrewda: Discussion in gitter regarding this. It messes up the gitter formatting to limit lines to 80 characters.

plugins/explain.py
jayvdb: Because we have a task to use it, and it is better that the task is a markdown+jinja beginner task rather than a coding task.

jayvdb: However @andrewda please remove this `if` here; it can be done in the template using jinja `if` logic.

This was requested previously at https://github.com/coala/corobo/pull/471#discussion_r154850103

plugins/labhub.py
```diff
@@ -129,10 +110,18 @@ def invite_cmd(self, msg, match):
 
             # send the invite
             self.TEAMS[team_mapping[team.lower()]].invite(invitee)
-            return self.INVITE_SUCCESS[team.lower()].format(invitee)
+            return tenv().get_template(
+                'promotions/{}.jinja2.md'.format(team.lower())
+            ).render(
+                username=invitee,
```
jayvdb: this should be `target=` so we have consistent variable names across templates.

plugins/labhub.py
```diff
@@ -129,10 +110,18 @@ def invite_cmd(self, msg, match):
...
+            )
         else:
-            return ('@{}, you are not a maintainer, only maintainers can invite'
-                    ' other people. Nice try :poop:'.format(inviter))
+            return tenv().get_template(
+                'errors/not-maintainer.jinja2.md'
+            ).render(
+                action='invite other people',
+                username=invitee,
```
jayvdb: `target=` , or maybe `target_username` or `subject_username` (just ideas; whatever is chosen should be consistently used in all template invocations)

plugins/lmgtfy.py
```diff
@@ -7,21 +7,10 @@ class Lmgtfy(BotPlugin):
...
-
     @re_botcmd(pattern=r'lmgtfy\s+(.+)',
-               re_cmd_name_help='lmgtfy <search-string>')
+               re_cmd_name_help='lmgtfy <search-string>',
+               template='lmgtfy.jinja2')
     def lmgtfy(self, msg, match):
         """I'm lazy, please google for me."""  # Ignore QuotesBear
         link = 'https://www.lmgtfy.com/?q=' + match.group(1)
-        return self.MSG.format(link)
+        return {'link': link}
```
jayvdb: pass in `match.group(1)`, let the template build the link, so it could be changed without code changes.

Also, that would mean that this plugin becomes very similar in purpose and structure to `explain` , only this plugin is for terms which are not (yet) part of the FAQ, overlapping a little with `explain google` in its intention.

jayvdb:
fyi, waiting for a potential newcomer fix to arrive; due to expire in a few hours, which I suspect will work nicely with Andrew's timezone anyway.

jayvdb:
unack https://github.com/coala/corobo/pull/471/commits/abc438afa3e284333a58ae6d21d9af3ddd99c21b

plugins/labhub.py
```diff
@@ -165,9 +159,11 @@ def create_issue_cmd(self, msg, match):
             iss = repo.create_issue(iss_title, iss_description + extra_msg)
             return 'Here you go: {}'.format(iss.url)
         else:
-            return ('Can\'t create an issue for a repository that does not '
-                    'exist. Please ensure that the repository is available '
-                    'and owned by the org.')
+            return tenv().get_template(
+                'errors/no-repository.jinja2.md'
```
jayvdb: trailing comma

jayvdb:
> I think it is better to have templates stored as plugins/templates/<plugin_name>/**.jinja2.md.

I'm flexible on this.  fwiw, I think we'll end up reusing some of labhub error message in other plugins.

plugins/explain.py
andrewda: This will fail a test because the user won't be mentioned. Should I do some Jinja2 logic to keep the test passing?

andrewda:
Will get this done in a few hours! Thanks for the reviews!

andrewda:
@jayvdb What's blocking this atm?

jayvdb:
#329

jayvdb:
ack https://github.com/coala/corobo/pull/471/commits/11ce167b47c96c3536e1de68a1b27e410353025c

PR #477, opened 2017-12-10
Meet/security hardening

meetmangukiya:
# Reviewers Checklist

### DIFF
--- a/plugins/labhub.py
+++ b/plugins/labhub.py
@@ -16,22 +16,31 @@ from utils.backends import message_link
 class LabHub(BotPlugin):
     """GitHub and GitLab utilities"""  # Ignore QuotesBear
 
+    HELLO_WORLD_MSG = ('Welcome @{}! :tada:\n\n'
+                       'Please go through https://coala.io/newcomer (skip '
+                       'instructions about using corobo). Find your first '
+                       'issue and ask one of the maintainers to assign you to '
+                       'it and we\'ll invite you to the org. Most issues will '
+                       'be explained there and in linked pages - it will save '
+                       'you a lot of time, just read it. *Really.*\n\n'
+                       '*Do not take an issue if you don\'t understand it on '
+                       'your own.*Especially if you are new you have to be '
+                       'aware that getting started with an open source '
+                       'community is not trivial: you will have to work hard '
+                       'and most likely become a better coder than you are '
+                       'now just as we all did.\n\nDon\'t get us wrong: we are '
+                       '*very* glad to have you with us on this journey into '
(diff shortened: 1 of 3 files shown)


plugins/labhub.py
```diff
@@ -143,7 +137,6 @@ def callback_message(self, msg):
                 # send the invite
                 self.send(msg.frm,
                           self.INVITE_SUCCESS['newcomers'].format(user))
-                self.TEAMS[self.GH_ORG_NAME + ' newcomers'].invite(user)
                 self.invited_users.add(user)
```
Vamshi99: Remove this line, also change regex such that it will not respond to `invite me`

tests/labhub_test.py
```diff
@@ -326,7 +325,7 @@ def test_invite_me(self):
 
         plugins.labhub.os.environ['GH_TOKEN'] = 'patched?'
         testbot.assertCommand('!invite me',
```
Vamshi99: This should be `hello world`

plugins/labhub.py
meetmangukiya: 1. Would remove the regex 👍 
2. Would change `self.invited_users` to `self.helloed_users`, since it is just tracking if it needs to trigger again or not

tests/labhub_test.py
meetmangukiya: ughh. we don't need this test case at all

plugins/labhub.py
Vamshi99: Also, when maintainer invites, invite message should be different right? as we changed the message

plugins/labhub.py
```diff
@@ -16,22 +16,31 @@
...
-                     '*Do not take an issue if you don\'t understand it on '
-                     'your own.*Especially if you are new you have to be aware '
-                     'that getting started with an open source community is '
-                     'not trivial: you will have to work hard and most likely '
-                     'become a better coder than you are now just as we all '
-                     'did.\n\nDon\'t get us wrong: we are *very* glad to have '
-                     'you with us on this journey into open source! We will '
-                     'also be there for you at all times to help you with '
-                     'actual problems. :)',
+        'newcomers': '@{} We\'ve just sent you an invite via email. Please '
```
Vamshi99: Why are you changing the welcome message, it should be same

jayvdb:
Please split into several PRs , addressing individual issues, so we can merge them quickly.
Needs a rebase.

jayvdb:
unack https://github.com/coala/corobo/pull/477/commits/f2e7a965ca0a12a83016fe1623ffaa3094dcae8c

Issue #371, opened 2017-11-24
Rename COBOT_ROOT to BOT_ROOT


jayvdb:
The env vars should be consistent with the naming system used by err, and the name "cobot" needs to die.

Need backwards compatibility for `COBOT_ROOT`, with a warning, as it will be removed after people have had a chance to change their scripts.

EmilyOng:
Hello, I would like to work on this but could you point me to the direction what this statement means and how i should go about doing it?

> Need backwards compatibility for COBOT_ROOT, with a warning, as it will be removed after people have had a chance to change their scripts.

Thanks 😄

yukiisbored:
Unassigned per request.

EverWinter23:
Would like to work on it.

PR #500, opened 2018-01-26
Rename COBOT_ROOT to BOT_ROOT

EverWinter23:
Add backward compatibilty for COBOT_ROOT with a
warning.

Closes https://github.com/coala/corobo/issues/371

# Reviewers Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Read more about what `corobo` could do for you
 
 ## Environment Variables
 
-1. `COBOT_ROOT` - absolute path of the project root.
+1. `BOT_ROOT` - absolute path of the project root.
 2. `BOT_PREFIX` - prefix to use for issuing bot commands.
 3. `BOT_ADMINS` - Admins of the errbot instance.
 4. `ROOMS` - Space separated list of rooms to join on startup. e.g.
--- a/config.py
+++ b/config.py
@@ -1,7 +1,15 @@
 import logging
 import os
 
-COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
+BOT_ROOT_KEY = 'BOT_ROOT'
+
+if 'COBOT_ROOT' in os.environ and BOT_ROOT_KEY not in os.environ:
+    logging.warning(
+        "Environment variable COBOT_ROOT is deprecated, use {} instead."
+        .format(BOT_ROOT_KEY))
+    BOT_ROOT_KEY = 'COBOT_ROOT'
+BOT_ROOT = os.environ.get(BOT_ROOT_KEY, os.getcwd())
+
 
 _BOT_IDENTITY_KEYS = (
     'endpoint',
@@ -34,7 +42,7 @@ if not BACKEND:
         BACKEND = 'Text'
 
 if BACKEND == 'Gitter':
-    BOT_EXTRA_BACKEND_DIR = os.path.join(COBOT_ROOT, 'err-backend-gitter')
+    BOT_EXTRA_BACKEND_DIR = os.path.join(BOT_ROOT, 'err-backend-gitter')
 else:
     BOT_EXTRA_BACKEND_DIR = None
 
@@ -46,10 +54,10 @@ if BOT_EXTRA_BACKEND_DIR:
 
 HIDE_RESTRICTED_COMMANDS = True
 
-BOT_DATA_DIR = os.path.join(COBOT_ROOT, 'data')
-BOT_EXTRA_PLUGIN_DIR = COBOT_ROOT
+BOT_DATA_DIR = os.path.join(BOT_ROOT, 'data')
+BOT_EXTRA_PLUGIN_DIR = BOT_ROOT
 
-BOT_LOG_FILE = os.path.join(COBOT_ROOT, 'errbot.log')
+BOT_LOG_FILE = os.path.join(BOT_ROOT, 'errbot.log')
 BOT_LOG_LEVEL = logging.DEBUG
 
 BOT_PREFIX = os.environ.get('BOT_PREFIX', 'corobo ')
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM python:3-alpine
 
-ENV COBOT_ROOT=/opt/errbot
+ENV BOT_ROOT=/opt/errbot
 
-ADD requirements.txt $COBOT_ROOT/requirements.txt
+ADD requirements.txt $BOT_ROOT/requirements.txt
 
 RUN apk add --no-cache libffi openssl git \
     && apk add --no-cache --virtual .build-deps \
@@ -10,16 +10,16 @@ RUN apk add --no-cache libffi openssl git \
            libc-dev \
            libffi-dev \
            openssl-dev \
-    && pip install -r $COBOT_ROOT/requirements.txt \
+    && pip install -r $BOT_ROOT/requirements.txt \
     && pip install slackclient python-telegram-bot \
     && apk del .build-deps
 
-ADD . $COBOT_ROOT
+ADD . $BOT_ROOT
 
 RUN addgroup -S errbot \
-    && adduser -h $COBOT_ROOT -G errbot -S errbot \
-    && mkdir -p $COBOT_ROOT/data $COBOT_ROOT/plugins \
-    && chown -R errbot:errbot $COBOT_ROOT
+    && adduser -h $BOT_ROOT -G errbot -S errbot \
+    && mkdir -p $BOT_ROOT/data $BOT_ROOT/plugins \
+    && chown -R errbot:errbot $BOT_ROOT
 
 USER errbot
 WORKDIR /opt/errbot
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -17,7 +17,7 @@ class Utils(BotPlugin):
     @botcmd(admin_only=True)
     def sync(self, msg, arg):
         """Sync the repository from github."""  # Ignore QuotesBear
-        repo = git.Repo(self.bot_config.COBOT_ROOT)
+        repo = git.Repo(self.bot_config.BOT_ROOT)
         try:
             repo.remote('origin').pull('--rebase')
             yield 'Sync\'d successfully! :tada:'
@@ -27,14 +27,14 @@ class Utils(BotPlugin):
     @botcmd
     def get_head(self, msg, arg):
         """Yields the head commit."""  # Ignore QuotesBear
-        repo = git.Repo(self.bot_config.COBOT_ROOT)
+        repo = git.Repo(self.bot_config.BOT_ROOT)
         head = repo.commit('HEAD')
         yield '`{}`: {}'.format(head.hexsha, head.message)
 
     @botcmd(admin_only=True)
     def install_requirements(self, msg, arg):
         """Installs requirements"""  # Ignore QuotesBear
-        os.chdir(self.bot_config.COBOT_ROOT)
+        os.chdir(self.bot_config.BOT_ROOT)
         ran = run('pip install -r requirements.txt')
         yield 'installing requirements...'
         yield ran.stdout.read().decode('utf-8')


kriti21:
You have not been assigned yet, have you?

config.py
```diff
@@ -1,7 +1,14 @@
...
 
-COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
+# Deprecate COBOT_ROOT
+BOT_ROOT = os.environ.get('BOT_ROOT', os.getcwd())
+if 'BOT_ROOT' not in os.environ:
+    os.environ.setdefault('BOT_ROOT', 'COBOT_ROOT')
+    logging.warning()
+    logging.warning('Deprecation warning: environment variable {} is'
+                    + 'replaced by {}.'
+                    .format('COBOT_ROOT', 'BOT_ROOT'))
```
pareksha: Why are you using ``+`` in between the string? You can simply write ``'Deprecation warning: environment variable {} is replaced by {}.'.format(..)``. This would be better.

pareksha: And is there any need of using ``logging.warning()``? There's no need of empty warning I suppose as you are supplying the warning on line ``9``.

EverWinter23:
About the short-log, the tag is not mandatory when you're making changes to a bunch of files and the change is about one particular issue.

EverWinter23:
Made some changes.

config.py
```diff
@@ -1,7 +1,12 @@
-import logging
+from logging import warning
 import os
```
pareksha: There is no need to change ``import logging``. We generally import the whole logging module.

EverWinter23:
Seems like I squashed a lot of commits into one.

config.py
EverWinter23: The following does not fit in 80 columns.
    logging.warning('Deprecation warning: environment variable {} is replaced by {}.'
            .format('COBOT_ROOT', 'BOT_ROOT'))

pareksha: You can do ``logging.warning(
'Deprecation warning: environment variable {} is replaced by {}.'
.format('COBOT_ROOT', 'BOT_ROOT'))``

pareksha: And you do not need to change ``import logging`` to ``from logging import warning``

EverWinter23: got it.

EverWinter23:
The requested changes have been made.

EverWinter23:
@pareksha @kriti waiting for review. :grin:

config.py
```diff
@@ -1,7 +1,13 @@
 import logging
 import os
 
-COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
+# Deprecate COBOT_ROOT
+BOT_ROOT = os.environ.get('BOT_ROOT', os.getcwd())
+if 'BOT_ROOT' not in os.environ:
```
pareksha: You could change ``if 'BOT_ROOT' not in os.environ`` to ``if 'COBOT_ROOT' in os.environ`` so as to ensure that ``COBOT_ROOT`` is completely removed from the environment and not on only some places.

EverWinter23:
@pareksha I think they both achieve the same thing. Two ways to do one thing. Same thing is achieved by that piece of code too.

EverWinter23:
@Makman2 Review request.

README.md
```diff
@@ -30,7 +30,8 @@ Read more about what `corobo` could do for you
 
 ## Environment Variables
 
-1. `COBOT_ROOT` - absolute path of the project root.
+1. `COBOT_ROOT` - absolute path of the project root. - Will soon be deprecated.
+1. `BOT_ROOT` - absolute path of the project root.
```
Makman2: 1. The numbering is wrong ;)
2. "Will soon be depracated" - your commit already deprecates it, so it is deprecated right now :)
3. I would suggest you combine those to
   ```
   `BOT_ROOT` - absolute path of the project root. (Previously `COBOT_ROOT`, which is now deprecated)
   ```
   or similar phrasing :)

config.py
```diff
@@ -1,7 +1,13 @@
 import logging
 import os
 
-COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
+# Deprecate COBOT_ROOT
+BOT_ROOT = os.environ.get('BOT_ROOT', os.getcwd())
+if 'COBOT_ROOT' in os.environ:
+    os.environ.setdefault('BOT_ROOT', 'COBOT_ROOT')
```
Makman2: this will never kick-in, right? `BOT_ROOT` will always have a value, and `os.environ['BOT_ROOT']` is never executed again.

I think also a warning message should be only emitted if `BOT_ROOT` wasn't set, but only `COBOT_ROOT`.

Makman2:
Commit message: I think there's no verb like "to deprecate sth. to sth.", you can just "deprecate sth."
Maybe do in the commit message
```
Deprecate COBOT_ROOT

... and use BOT_ROOT instead. A warning is issued if `COBOT_ROOT` is used regardless.

Closes xxx
```

Makman2:
unack 30ac362fe473bccd05a122a88a77be80a0e8128c

config.py
EverWinter23: The 6th line will kick in if 'COBOT_ROOT' env var is set and will change this env var to 'BOT_ROOT'.
The warning message will be emitted if instance uses 'COBOT_ROOT' instead of 'BOT_ROOT'.

README.md
```diff
@@ -30,7 +30,8 @@ Read more about what `corobo` could do for you
 
 ## Environment Variables
 
-1. `COBOT_ROOT` - absolute path of the project root.
+1. `BOT_ROOT` - absolute path of the project root. (Previously `COBOT_ROOT`, 
+    which is now deprecated)
```
Makman2: Missing full-stop in sentence inside braces.

config.py
Makman2: Nah there's something wrong. 1st: You are setting the string `'COBOT_ROOT'` itself for the `BOT_ROOT` key inside the environment variables. 2nd: No code will ever call `os.environ['BOT_ROOT']` again, so this is meaningless.

So I played through all cases:
1. Only `BOT_ROOT` was set in env: `BOT_ROOT` var gets set to the env-var value.
2. Only `COBOT_ROOT` was set in env: `BOT_ROOT` is set to `os.getcwd()`, the `BOT_ROOT` key inside env gets the value `'COBOT_ROOT'` (a string) and issues a deprecation warning
3. Both were set in env: `BOT_ROOT` and `COBOT_ROOT`: `BOT_ROOT` is set to env value, the `BOT_ROOT` key is set to `'COBOT_ROOT'` string in `os.environ` and a deprecation warning is issued.

I would only issue a deprecation warning if only `COBOT_ROOT` is present, but not `BOT_ROOT`. Because then the user has missed the env-var change. But if both are present, he might have it there on purpose.

Makman2:
unack 1672ca6b3518e1d229fb9370302c6b9dfd1b61b8

EverWinter23:
@Makman2 Done.

Makman2:
@EverWinter23 please see https://github.com/coala/corobo/pull/500/commits/7600eceb37444ed36ae48c4405b24faba1898244#r168955621 :)

EverWinter23:
@Makman2 Changes done.

config.py
```diff
@@ -1,7 +1,13 @@
 import logging
 import os
 
-COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
+# Deprecate COBOT_ROOT
+if 'COBOT_ROOT' in os.environ and 'BOT_ROOT' not in os.environ:
+    logging.warning(
+        'Deprecation warning: environment variable {} is replaced by {}.'
```
Makman2: --> "Environment variable {} is deprecated, use {} instead."

For "Deprecation *warning*", the word "warning" would occur twice in the same log entry, because you also log with `logging.warning` ;)

config.py
```diff
@@ -1,7 +1,13 @@
...
+        .format('COBOT_ROOT', 'BOT_ROOT'))
+BOT_ROOT = os.environ.get('BOT_ROOT', os.getcwd())
```
Makman2: While we're on it, you maybe want to add another variable `BOT_ROOT_KEY = 'BOT_ROOT'`, then you don't have to retype the string in all used locations :)

Makman2:
unack 64f7a346ab1b01afa1fa3603ce3174491cc6f0d9

EverWinter23:
Done

Makman2:
ack cce8764c1f3b0a710320b95dfcdc2e9f1956da84

Makman2:
Do we have build problems on master currently?

Vamshi99:
@EverWinter23 rebase.

Makman2:
@gitmate-bot rebase

EverWinter23:
@Makman2 Why is the build failing on semaphoreci??

EverWinter23:
@Makman2 Build Passed.

config.py
```diff
@@ -1,7 +1,15 @@
...
 
-COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
+BOT_ROOT_KEY = 'BOT_ROOT'
+
+# Deprecate COBOT_ROOT
+if 'COBOT_ROOT' in os.environ and BOT_ROOT_KEY not in os.environ:
+    logging.warning(
+        'Environment variable {} is replaced by {}.'
+        .format('COBOT_ROOT', BOT_ROOT_KEY))
+BOT_ROOT = os.environ.get(BOT_ROOT_KEY, os.getcwd())
```
jayvdb: if `COBOT_ROOT` was set, you dont load it.

jayvdb:
unack https://github.com/coala/corobo/pull/500/commits/9240dccf61b226ffd3b2046e5dd71a393f8044d5

config.py
EverWinter23: Which line are you referring to? The 14th one? Or the warning?

Makman2: The warning. You just warn that it's deprecated, but don't use the old env-variable completely.

I would recommend to pick my message from an earlier comment so people know that the variable is still functional:
> "Environment variable {} is deprecated, use {} instead."

EverWinter23: :+1:

EverWinter23:
@Makman2 Done.

EverWinter23:
@Makman2 Pls review

config.py
```diff
@@ -1,7 +1,17 @@
...
 
-COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
+BOT_ROOT_KEY = 'BOT_ROOT'
+
+# Deprecate COBOT_ROOT
+if 'COBOT_ROOT' in os.environ and BOT_ROOT_KEY not in os.environ:
+    logging.warning(
+        "Environment variable {} is deprecated, use {} instead."
+        .format('COBOT_ROOT', BOT_ROOT_KEY))
+    # Add support for COBOT_ROOT
```
Makman2: Add support for COBOT_ROOT?

Makman2: Sorry I should have been more precise, I mean the Python comment itself. I see that it's deprecated when looking at the message above, this Python comment just confuses me^^ I directly asked myself "What additional support is being offered here besides handling the deprecation?", as `COBOT_ROOT` was already "supported" prior to this change.

config.py
```diff
@@ -1,7 +1,17 @@
...
 import os
 
-COBOT_ROOT = os.environ.get('COBOT_ROOT', os.getcwd())
+BOT_ROOT_KEY = 'BOT_ROOT'
+
+# Deprecate COBOT_ROOT
+if 'COBOT_ROOT' in os.environ and BOT_ROOT_KEY not in os.environ:
+    logging.warning(
+        "Environment variable {} is deprecated, use {} instead."
+        .format('COBOT_ROOT', BOT_ROOT_KEY))
```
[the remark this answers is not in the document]
EverWinter23: Will make changes.

EverWinter23:
@nvzard got it.

EverWinter23:
@Makman2 Review

Makman2:
ack d1223b76dc0528fa7765d025f83d6f2fc367400d

Makman2:
@gitmate-bot ff

Issue #307, opened 2017-10-21
Add “invite me” in corobo help


prnvdixit:
The “corobo help” command should list “invite me” too with a proper description.

Label: initiatives/gci

jayvdb:
Note the help is found associated with the `invite_cmd` in `labhub.py` , in the `re_cmd_name_help`

random-access7:
Hello, I would really like to work on this. If possible can I be assigned this task?

random-access7:
Also, I was going through the script [labhub.py](https://github.com/coala/corobo/blob/master/plugins/labhub.py) and for the `invite_cmd` method, I see that the regex can match to "invite @userto developers" while `match.group(1)` will return "user" (instead of "userto"). I guess this wouldn't affect the functionality much, but might just be a corner-case error? If possible, could I work on this and [#307](https://github.com/coala/corobo/issues/307)?

Makman2:
Please file a new issue, such discussions shouldn't clutter unrelated issues ;)

PR #510, opened 2018-02-23
Updating file with respect to #307

random-access7:
The 'corobo help' will now show how the invite command can work in two ways. The two formats, as evident from the regex pattern can be understood as: 
1. The first can be in the format 'invite @username to developers' with the regex matching group(1) to username and group(2) to developers. However, if the format is 'invite @username', the regex will make group(2) to None, which is taken as 'newcomers' by default within the method.
2. The second way is to use 'invite me' which will send the invite to the username who originally wrote 'invite me' and the group will be taken as 'newcomers' by default.

# Reviewers Checklist


Makman2:
Please don't close PRs, such issues can be always addressed in the same PR.

random-access7:
I apologize for the mistake, @Makman2 . Will keep this in mind next time.

PR #568, opened 2018-06-21
Moban

jayvdb:
# Reviewers Checklist

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,6 +1,21 @@
-pytest
-coverage
-pytest-cov
-requests_mock
-codecov
+coverage~=4.3.4
+coverage-env-plugin~=0.1
+coverage-config-reload-plugin~=0.2
+codecov~=2.0.5
+moban~=0.0.9
+packaging~=16.8
+pytest~=3.6.1
+pytest-cov~=2.4
+pytest-env~=0.6.0
+pytest-error-for-skips~=1.0
+pytest-instafail~=0.3.0
+pytest-mock~=1.1
+pytest-profiling~=1.3.0
+pytest-reorder~=0.1.0
+pytest-reqs~=0.0.6
+pytest-timeout~=1.3.0
+pytest-xdist~=1.15
+requests-mock~=1.2
+pip!=9.0.2, !=10.0.*
+wheel~=0.29
 vcrpy
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,8 @@
-*.orig
+# moban hashes
+.moban.hashes
+
+# Extra rules from https://github.com/github/gitignore/
+# Python rules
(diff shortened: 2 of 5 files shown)


shikharvaish28:
@jayvdb https://github.com/coala/corobo/pull/568/commits/c37af69bea7ad79a5dc7014b060a8f5bf6e19181 looks fine to me.

Just out of curiosity, what was intended in https://github.com/coala/corobo/pull/568/commits/d3e355f6a80b62644f187ae752f5ba8ef1cc90fa. The issue you mentioned in https://github.com/coala/corobo/pull/568/commits/d3e355f6a80b62644f187ae752f5ba8ef1cc90fa was closed but what are the other lines for.

jayvdb:
Hiya @shikharvaish28, thanks for reviewing.

We now have a template for creating the `.gitignore`, at https://gitlab.com/coala/mobans/blob/master/templates/gitignore.jj2 .

It pulls in a bunch of rules from https://github.com/github/gitignore .
The same template is being used on all of our repos, so that they are always in sync, and being updated whenever new rules are added to https://github.com/github/gitignore .

This also means that individual additions in each repo are no longer desirable; instead an addition needs to be proposed at https://github.com/github/gitignore , where they will be reviewed carefully and will benefit more people; when accepted, the change will flow through to our repos.

nvzard:
ah failed again due to that bugger

test-requirements.txt
```diff
@@ -1,6 +1,21 @@
...
+pytest-instafail~=0.3.0
+pytest-mock~=1.1
+pytest-profiling~=1.3.0
+pytest-reorder~=0.1.0
+pytest-reqs~=0.0.6
+pytest-timeout~=1.3.0
+pytest-xdist~=1.15
+requests-mock~=1.2
+pip!=9.0.2, !=10.0.*
+wheel~=0.29
```
nvzard: What is the need of all these extra `pytest-*` deps, isn't `pytest~=3.6.1` supposed to be enough?

jayvdb: These are all the lovely stuff which work on coala/coala repo, and the template used is being deployed to every other coala repo as well so we can use them there.
pip caching means the extra setup time is negligible.  And the plugins all should do nothing unless enabled in `setup.cfg`.

After this, you can enable profiling -- see https://github.com/coala/coala/pull/5548 , which also shows time spent in pytest plugins so enabling other features can be 'costed' and rejected if the plugin is too expensive.

nvzard: Woah :o Amazing !!

jayvdb:
ack https://github.com/coala/corobo/pull/568/commits/2d0c920a285263278d2b7f419055ffba59d18604

jayvdb:
@gitmate-bot ff

Issue #564, opened 2018-06-16
test_alive failure


jayvdb:
I've seen this a few times, but we dont seem to have an issue about it.

`TestLabHub.test_alive` fails occasionally.

e.g. https://travis-ci.org/coala/corobo/builds/392987950 for #563

Something about this test isnt resilient enough to be reliable.

jayvdb:
And again
https://travis-ci.org/coala/corobo/builds/394806698

jayvdb:
Failure on master https://travis-ci.org/coala/corobo/builds/394804787

PR #570, opened 2018-06-21
labhub_test: Increase timeout to fix random CI failures

nvzard:
Default 5 sec timeout was causing unexpected random CI
failure.

Closes https://github.com/coala/corobo/issues/564

# Reviewers Checklist

### DIFF
--- a/tests/labhub_test.py
+++ b/tests/labhub_test.py
@@ -400,7 +400,7 @@ class TestLabHub(unittest.TestCase):
             labhub.gh_repos['coala'].search_mrs.return_value = []
             testbot.assertCommand('!pr stats 5hours',
                                   '0 PRs opened in last 5 hours\n'
-                                  'The community is dead')
+                                  'The community is dead', timeout=100)
 
             labhub.gh_repos['coala'].search_mrs.return_value = [
                 1, 2, 3, 4, 5,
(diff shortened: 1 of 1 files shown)


nvzard:
Because IGitt is making more that one requests while calling `write_repositories` method with an invalid token. That is sometimes taking more than 5secs and so the bot message queue is showing a `queue.Empty` exception as the response is delayed due to IGitt's requests.

jayvdb:
fwiw, if this unbreaks the tree, I think it is an acceptable workaround *if* an issue is raised in `IGitt`so that followup work is done to prevent it biting us again.

nvzard:
IGitt is doing its work as it is supposed to but it is making a second trial request because the token is invalid. This portion is handled by backoff https://pypi.org/project/backoff/ which makes the request function retry until some condition is met.

jayvdb:
@nvzard , IGitt is causing this failure.  If you cant disable the 'feature' of `IGitt` causing this failure, then that is a problem in `IGitt`.  And please remove 'fix' from this commit message.  It hasnt fixed anything.  It has only obscured the problem, and it will bite us again another day.  This is a workaround at best. :P  A very useful workaround to get the builds working again, but we still need a solution to the problem.

jayvdb:
50 char limit on commit shortlog.

jayvdb:
ack https://github.com/coala/corobo/commit/2feedd67eac2af77c47ff44f1c6ff7a3c651615c

jayvdb:
@gitmate-bot ff

Issue #586, opened 2018-07-20
Implement DefaultConfigurationMixin


nvzard:
`DefaultConfigurationMixin` will allow configuration of plugins from `config.py` so that admins won't have to configure the plugin every time the bot is deployed.

PR #588, opened 2018-07-20
configuration: Use default configurations

nvzard:
Closes https://github.com/coala/corobo/issues/586
Closes https://github.com/coala/corobo/issues/587

### DIFF
--- a/plugins/spam.py
+++ b/plugins/spam.py
@@ -1,12 +1,11 @@
-from itertools import chain
-
 from errbot import BotPlugin
 from errbot.templating import tenv
 
 from plugins import constants
+from utils.mixin import DefaultConfigMixin
 
 
-class SpammingAlert(BotPlugin):
+class SpammingAlert(DefaultConfigMixin, BotPlugin):
     """
     A plugin which alerts the user that they might be spamming.
     """
@@ -15,21 +14,6 @@ class SpammingAlert(BotPlugin):
         'MAX_MSG_LEN': constants.MAX_MSG_LEN,
         'MAX_LINES': constants.MAX_LINES
     }
-
-    def get_configuration_template(self):
-        return self.CONFIG_TEMPLATE
-
-    def configure(self, configuration):
-        """
-        Enable partial configuration.
-        """
-
-        if configuration:
-            config = dict(chain(self.CONFIG_TEMPLATE.items(),
-                                configuration.items()))
-        else:
-            config = self.CONFIG_TEMPLATE
-        super(SpammingAlert, self).configure(config)
 
     def check_configuration(self, configuration):
         pass
--- a/tests/spam_test.py
+++ b/tests/spam_test.py
@@ -1,4 +1,5 @@
 import logging
+import queue
 import unittest
 
 from errbot.backends.test import TestBot
@@ -11,6 +12,9 @@ class TestSpam(unittest.TestCase):
                                loglevel=logging.ERROR)
         self.testbot.start()
 
+    def tearDown(self):
+        self.testbot.stop()
+
     def test_spam_callback(self):
         self.testbot.assertCommand('c'*1001, 'you\'re spamming')
         self.testbot.assertCommand('\n\n'*11, 'you\'re spamming')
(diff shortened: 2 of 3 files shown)


config.py
```diff
@@ -107,3 +107,10 @@
     'LabHub:*': {'allowprivate': False}}
 
 AUTOINSTALL_DEPS = True
+
+DEFAULT_CONFIG = {
+    'SpammingAlert': {
+        'MAX_MSG_LEN': 1000,
+        'MAX_LINES': 20,
+        },
```
meetmangukiya: one level less of indentation, same for next line.

plugins/spam.py
```diff
@@ -3,32 +3,28 @@
...
     """
     A plugin which alerts the user that they might be spamming.
     """
 
-    CONFIG_TEMPLATE = {
-        'MAX_MSG_LEN': constants.MAX_MSG_LEN,
-        'MAX_LINES': constants.MAX_LINES
-    }
-
     def get_configuration_template(self):
```
meetmangukiya: Reorder class definition as `SpammingAlert(DefaultConfigMixin, BotPlugin)` and you can eliminate the `get_configuration_template` overriding completely.

utils/mixin.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self, bot, name=None):
+        super().__init__(bot, name=name)
+        default_config = self._default_config
+        if default_config and not self.config:
+            super().configure(default_config)
+
+    def get_configuration_template(self):
+        default_config = self._default_config
+        if default_config:
+            return default_config
```
meetmangukiya: create a `configure` method that does something like:
```py
def configure(self, config):
  new = self.bot_config.DEFAULT_CONFIG[self.name]
  for key, value in config:
    new[key] = value
  self.config = new

**Comment by nvzard** [comment]
Since the `SpammingAlert` commit is dependent on the `Mixin` commit. I pushed the commits it in the same PR.
Plus, I thought `DefaultConfigMixin` PR without any adaptation shown would be bland.
>Unrelated commits = Different PRs

File: utils/mixin.py [hunk 4]
```diff
@@ -0,0 +1,24 @@
...
+    @property
+    def _default_config(self):
+        if (self.bot_config.DEFAULT_CONFIG and self.name
+            in self.bot_config.DEFAULT_CONFIG):
+            return self.bot_config.DEFAULT_CONFIG[self.name]
+
+    def __init__(self, bot, name=None):
+        super().__init__(bot, name=name)
+        default_config = self._default_config
+        if default_config:
```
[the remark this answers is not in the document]
jayvdb: No it doesnt.

It makes *both* classes the base class, irrespective of order.

What changes is which is called first, and the mixin must be first.

Then you call the super to let `BotPlugin` do its work at beginning or end of the mixin methods depending on the needs of the mixin.

plugins/spam.py
```diff
@@ -3,32 +3,25 @@
...
+from utils.mixin import DefaultConfigMixin
 
 
-class SpammingAlert(BotPlugin):
+class SpammingAlert(DefaultConfigMixin, BotPlugin):
     """
     A plugin which alerts the user that they might be spamming.
     """
 
-    CONFIG_TEMPLATE = {
```
jayvdb: what the??

a plugin should include its own config template in its own source file, and not *rely* on config.py.

the plugin config information stays here in the plugin.

`config.py` should only have *overrides* for the orgs preferred starting config, only where it differs from the bot defaults.

otherwise it is mandatory to include in `config.py` a bunch of plugin config values which are not coala specific.

and if someone copies only the plugin file , they should be ok.  (they will also need to copy `DefaultConfigMixin` , which is annoying but we can solve that later.)  `constants.py` should go away eventually when each plugin is able to be customised by orgs easily.

utils/mixin.py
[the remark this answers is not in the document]
nvzard: Got it :+1:
It took me more than the required time to understand this but I finally understood now :sweat_smile: .

I got confused by the Right->Left sequence mentioned here https://www.ianlewis.org/en/mixins-and-python and ignored the fact that `__init__` will be called based on the left to right sequence.

Earlier `__init__` of `DefaultConfigMixin` with sequece `(BotPlugin, DefaultConfigMixin)` was not even being called.

Thanks for explaining.

utils/mixin.py
```diff
@@ -0,0 +1,38 @@
+from itertools import chain
+
+class DefaultConfigMixin():
+
+    @property
+    def _default_config(self):
+        if (self.bot_config.DEFAULT_CONFIG and self.name
+            in self.bot_config.DEFAULT_CONFIG):
```
jayvdb: lines with an `if` clause should be indented four additional spaces so that do not align with the statements in the indented block

tests/spam_test.py
```diff
@@ -7,8 +7,17 @@
...
+                'SpammingAlert': {
+                    'MAX_MSG_LEN': 1000,
+                    'MAX_LINES': 10,
+                }
+            }
+        }
         self.testbot = TestBot(extra_plugin_dir='plugins',
-                               loglevel=logging.ERROR)
+                               loglevel=logging.ERROR,
+                               extra_config=extra_config)
```
jayvdb: extra config here should override the plugin defaults, so these tests passing show that your code is not working correctly.

Anyway, dont add `extra_config` here.  This test class is good without it.

Create an additional test class `TestSpamExtraConfig` which verifies the extra config overrides the plugin defaults.

nvzard: done

tests/spam_test.py
```diff
@@ -22,3 +22,32 @@ def test_spam_configuration(self):
...
+                }
+            }
+        }
+        self.testbot = TestBot(extra_plugin_dir='plugins',
+                               loglevel=logging.ERROR,
+                               extra_config=extra_config)
+        self.testbot.start()
+
+    def test_spam_extra_config_callback(self):
+        self.testbot.assertCommand('c'*5001, 'you\'re spamming')
```
jayvdb: this does not confirm that the extra config is being used, because 5001 is above the default of `1000` and the extra config value of `500`.  So it will only verify that the command works; it doesnt verify that the command is triggered only if above the number of lines setting.

An obvious omission from this test module is asserting that a person can post content below these sizes and no errors occur.  I'm not sure whether errbot test framework has a suitable helper for that.

the next line is also not 100% ok.   `'\n\n'*6` is 12 new line characters, and that is lower than `constants.MAX_LINES` -- so, all good?

not quite ... it assumes that pytest did not run `test_spam_extra_config_configuration` before `test_spam_extra_config_callback`.
There is no guarantee of that.
https://duckduckgo.com/?q=pytest+plugin+random+order

Tests should not leave the test bed modified after they have finished

Simplest fix is to add a `tearDown` if using `unittest.TestCase` to reset the config to the default state, or ... less correct .. (re)set to a constant state in `setUp`.

This also needs to be done to `TestSpam`.

nvzard: Shit. That 5001 was a typo. It was supposed to be 501. :sweat_smile:

tests/spam_test.py
```diff
@@ -19,6 +23,50 @@ def test_spam_configuration(self):
         self.testbot.assertCommand('!plugin config SpammingAlert '
                                    '{\'MAX_LINES\': 10}',
                                    'configuration done')
+        self.testbot.assertCommand('!plugin config SpammingAlert',
+                                   '{\'MAX_LINES\': 10}')
+        self.testbot.assertCommand('!plugin config SpammingAlert '
+                                   '{\'MAX_LINES\': 20, \'MAX_MSG_LEN\': 200}',
```
jayvdb: use `'{\'MAX_MSG_LEN\':1, \'MAX_LINES\': 1}',` from the original code, so that the diff is aligned properly.

nvzard: @jayvdb this sets `MAX_MSG_LEN = 1`. So when the command msg `!plugin config SpammingAlert` is sent to the bot in the next line. The takes the cmd msg as a spam since MAX_MSG_LEN is set to 1.

config.py
```diff
@@ -111,3 +111,10 @@
     'LabHub:*': {'allowprivate': False}}
 
 AUTOINSTALL_DEPS = True
+
+DEFAULT_CONFIG = {
+    'SpammingAlert': {
```
jayvdb: Please do not put any overrides here.

These are the same as the bot defaults, so unnecessary.

jayvdb:
ack 7179622

jayvdb:
@gitmate-bot ff

jayvdb:
@gitmate-bot ff

Issue #569, opened 2018-06-21
Remove need for the empty data subdirectory


jayvdb:
`/data/` should be created at runtime if it is missing.

jayvdb:
cross linking with https://gitlab.com/coala/mobans/issues/48

jayvdb:
This *may* be an upstream fix.

nvzard:
https://github.com/errbotio/errbot/pull/1232

nvzard:
@jayvdb https://github.com/errbotio/errbot/pull/1232#issuecomment-400090465

PR #591, opened 2018-07-23
config.py: Create empty data sub-directory

nvzard:
Remove need for the empty `data/` subdirectory.

Closes https://github.com/coala/corobo/issues/569

### DIFF
--- a/config.py
+++ b/config.py
@@ -55,6 +55,10 @@ if BOT_EXTRA_BACKEND_DIR:
 HIDE_RESTRICTED_COMMANDS = True
 
 BOT_DATA_DIR = os.path.join(BOT_ROOT, 'data')
+if not os.path.isdir(BOT_DATA_DIR):
+    # create an empty data directory
+    os.mkdir(BOT_DATA_DIR)
+
 BOT_EXTRA_PLUGIN_DIR = BOT_ROOT
 
 BOT_LOG_FILE = os.path.join(BOT_ROOT, 'errbot.log')
--- a/data/.gitignore
+++ b/data/.gitignore
@@ -1,2 +1,0 @@
-*
-!.gitignore


config.py
```diff
@@ -54,6 +54,16 @@
...
+
+def make_dir(path):
+    try:
+        os.mkdir(os.path.join(BOT_ROOT, path))
+    except OSError:
+        if os.path.isdir(os.path.join(BOT_ROOT, path)):
+            pass
+
+
+make_dir('data')
```
jayvdb: you should use `BOT_DATA_DIR`

config.py
```diff
@@ -54,6 +54,16 @@
...
+
+def make_dir(path):
+    try:
+        os.mkdir(os.path.join(BOT_ROOT, path))
+    except OSError:
+        if os.path.isdir(os.path.join(BOT_ROOT, path)):
+            pass
+
+
+make_dir(BOT_DATA_DIR)
```
jayvdb: wtf.

This var is being set below

jayvdb: Also please add a note in the log that the directory is being created.

config.py
```diff
@@ -54,7 +54,17 @@
 
 HIDE_RESTRICTED_COMMANDS = True
 
+
+def make_dir(path):
+    try:
+        os.mkdir(os.path.join(BOT_ROOT, path))
+    except OSError:
+        if os.path.isdir(os.path.join(BOT_ROOT, path)):
```
jayvdb: why not do this first before trying to make the directory?

will `os.mkdir` recursively create subdirectories that are missing?

nvzard: I removed the method completely and just did
```
if os.path.isdir(os.path.join(BOT_ROOT, BOT_DATA_DIR)):
    os.mkdir(os.path.join(BOT_ROOT, BOT_DATA_DIR))
```

jayvdb:
ack https://github.com/coala/corobo/pull/591/commits/4c8bb8331b7869fc14d0f8a004a9020a177d52ef

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

Issue #441, opened 2017-12-01
Repeated invite message 


co-robo:
Please refer to https://gitter.im/coala/coala?at=5a2198e9540c78242d7146a9 for more information 
difficulty/newcomer
Opened by @RaiVaibhav at [Gitter](https://gitter.im/coala/coala?at=5a219a8d540c78242d715177)

jayvdb:
Very likely due to the subclassing in https://github.com/coala/corobo/commit/07fb758f046a15def302f4f3b00408e352bc620b

Makman2:
@jayvdb not sure if we should revert the commit for now... I've seen that even issues are filed twice^^

meetmangukiya:
The most affected will be the create issues, since it'll create two issues. I guess we should revert it if nobody gets around fixing it for another day, we don't want dupes for every new issue being filed. Else I might fix it tomorrow and in that case we should have it reverted.

jayvdb:
subclassing was a bad idea.  That still means two instances of `LabHub` are fetching everything.  But this is our problem - we recommended the student do this approach, we ack'd it, and I merged it.
Working on a solution now.

PR #598, opened 2018-08-04
GitStats: Remove LabHub inheritance

nvzard:
Inheritence was resulting in labhub commands
being triggered twice when used. This creates
dependency on labhub instead of inheritance.

Fixes https://github.com/coala/corobo/issues/441


jayvdb:
ack 0caa68b

jayvdb:
@gitmate-bot ff

Issue #358, opened 2017-11-22
Limit pitchfork command to developers


jayvdb:
https://github.com/coala/corobo/issues/36 requested "When issued by an authorized person to cobot,.." , however that bit wasnt done.
This needs to be fixed before GCI or there will be mayhem when this easter egg is found.

aashraybhandar1:
@jayvdb  How do I check if the person issuing the command is a developer? Do I have to use labhub to do the same?

aashraybhandar1:
Also should pitchfork be part of labhub ?? @jayvdb

jayvdb:
Yes, the data you need is in labhub, so the pitchfork plugin needs to *depend on* labhub, and get the data from there.  pitchfork should not be part of labhub

Vamshi99:
ping @aashraybhandar1 Are you working on this? If not, let us know.

aashraybhandar1:
Hey I'm really sorry.I had my end semester exams so wasn't able to continue working on it. I do want to work on this issue and I will try to wrap it up in the and atleast send the first pull request in the next couple of days.

jayvdb:
Turns out this is hard due to the problem of testing dependent plugins

jayvdb:
This needs to be implemented as a filter to prevent access to any command based on configuration.

PR #577, opened 2018-07-04
LabHub: Use activate and configuration templates

nvzard:
`__init__` is executed at load time and
and any issue in the plugin will lead
to it not showing up.
LifeCycle: `__init__`->configured->enable

Adapts to the `DefaultConfigMixin` and
migrated to using configuration templates.

Closes https://github.com/coala/corobo/issues/554
Closes https://github.com/coala/corobo/issues/382

PR #599, opened 2018-08-04
Pitchfork: Make pitchfork private to org members

nvzard:
Pitchfork command is accessible only by
members of the organization.

Closes https://github.com/coala/corobo/issues/358


plugins/pitchfork.plug
```diff
@@ -1,6 +1,7 @@
 [Core]
-name = pitchfork
+name = Pitchfork
```
jayvdb: why change name?

tests/pitchfork_test.py
```diff
@@ -1,9 +1,54 @@
...
+from unittest.mock import create_autospec
+from tests.corobo_test_case import CoroboTestCase
 
-def test(testbot):
-    testbot.assertCommand('!pitchfork @meet', 'being pitchforked')
-    testbot.assertCommand('!pitchfork @meet down to hell', 'being pitchforked')
-    testbot.assertCommand('!pitchfork meet to hell', 'being pitchforked')
-    testbot.assertCommand('!pitchfork', 'Usage')
+
+class TestPitchfork(CoroboTestCase):
```
jayvdb: Many of our plugins which be changed to using labhub permissions.

create a LabHubTestCase subclassing CoroboTestCase which does all the setUp necessary for testing a plugin which depends on LabHub.

jayvdb:
All other plugins which should disabled: the_rules, wolfram_alpha, answer, lmgtfy, ghetto, explain, nevermind.

i.e. everything except searchdocs .

Also this is blocked by https://github.com/coala/corobo/pull/577 .

If LabHub is disabled, all of the corobo plugins (listed at https://github.com/errbotio/errbot/wiki ) should not become useless.

plugins/pitchfork.py
```diff
@@ -3,14 +3,20 @@
...
 
 
 class Pitchfork(BotPlugin):
     """
     To pitchfork users down to ...
     """
 
+    def activate(self):
+        super().activate()
+        self.team_mapping = self.get_plugin('LabHub').team_mapping
```
jayvdb: It shouldnt be necessary for all our plugins to have this duplicated `activate` code.

Again, creating a subclass or mixin for those plugins would be helpful.

plugins/pitchfork.plug
nvzard: To make it workable with CoroboTestCase. According to errbot documentation plugin name and the class name should be same.

plugins/pitchfork.py
jayvdb: not done

and also this shouldnt be using labhub directly.

labhub should be providing a filter

https://github.com/coala/corobo/pull/599#pullrequestreview-143397728

plugins/pitchfork.py
```diff
@@ -3,14 +3,20 @@
 import textwrap
 
 from errbot import BotPlugin, botcmd
+from plugins.labhub import members_only
```
jayvdb: we do not want to make everything dependant on labhub in order to prevent people from misusing all of the other commands.

plugins/labhub.py
```diff
@@ -108,6 +89,21 @@ def team_mapping(self):
             'maintainers': self.TEAMS[self.GH_ORG_NAME + ' maintainers'],
         }
 
+    @cmdfilter
+    def members_only(self, msg, cmd, args, dry_run):
+        user = msg.frm.nick
+        commands = ['assign_cmd', 'create_issue_cmd', 'invite_cmd', 'mark_cmd',
```
jayvdb: ideally this is configurable, but that can be done after gsoc.

put it in constants for now.

plugins/labhub.py
```diff
@@ -138,42 +133,40 @@ def invite_cmd(self, msg, match):
 
         valid_teams = self.team_mapping()
         if team not in valid_teams:
-            yield 'Please select from one of the valid teams: ' + ', '.join(
+            return 'Please select from one of the valid teams: ' + ', '.join(
```
jayvdb: yield is better than return.

plugins/labhub.py
```diff
@@ -108,6 +89,21 @@ def team_mapping(self):
...
+        user = msg.frm.nick
+        commands = ['assign_cmd', 'create_issue_cmd', 'invite_cmd', 'mark_cmd',
+                    'pr_stats', 'unassign_cmd', 'pitchfork', 'the_rules', 'wa',
+                    'answer', 'lmgtfy', 'ghetto', 'explain', 'nevermind']
+
+        if cmd in commands:
+            for team in self.team_mapping().values():
+                if team.is_member(user):
+                    return msg, cmd, args
+            return None, None, None
```
jayvdb: cant we also emit a message `'You need to be a member of this organization to use this command.'` so that the tests dont change?

perhaps change it to `return None, 'echo', 'You need to be a member of this organization to use this command.'` and add an `echo` command?

tests/labhub_test.py
```diff
@@ -180,13 +183,14 @@ def test_create_issue_cmd(self):
...
-            '''),
-            'You need to be a member of this organization to use this command.'
-        )
+        with self.assertRaises(queue.Empty):
+            testbot_public.assertCommand(
+                textwrap.dedent('''\
+                    !new issue repository this is the title
+                    body
+                '''),
+                'Here you go'
```
jayvdb: ideally this test doesnt change, but if it has to, change this to garbage so it is obvious it shouldnt work

tests/labhub_test.py
```diff
@@ -106,10 +106,13 @@ def test_invite_cmd(self):
...
+        self.mock_team.is_member.return_value = False
+        self.labhub.TEAMS = self.teams
 
-        testbot.assertCommand(
-            '!invite meet to newcomers',
-            'You need to be a member of this organization to use this command')
+        with self.assertRaises(queue.Empty):
+            testbot.assertCommand(
+                '!invite meet to newcomers',
+                ':poop:')
```
jayvdb: something more crazy than `:poop:` ; that still feels like it might be a legitimate response from the command.

plugins/labhub.py
nvzard: Okay, I'll take care of this.

tests/labhub_test.py
```diff
@@ -421,7 +422,7 @@ def test_mark_cmd(self):
         # not a member of org
         self.mock_team.is_member.return_value = False
         testbot.assertCommand(
-            cmd_github.format('pending review', 'coala', 'a', '23'),
+            cmd_github.format('pending review', 'coala', 'test', '23'),
```
jayvdb: why is this being changed?

plugins/labhub.py
jayvdb: Not done yet

Issue #601, opened 2018-08-05
plugins/coatils.py: Remove empty `__init__`


jayvdb:
The `__init__` in this module only invokes super, which will happen anyway if the `__init__` is removed.

This is medium importance because we do not want any plugin to be using `__init__`, as they should be using `activate`. (c.f. https://github.com/coala/corobo/pull/577)

PR #604, opened 2018-08-05
plugins/coatils.py: Remove __init__ constructor

bhavya17037:
This removes __init__ initializer as the __init__ in this module
only invokes super, which happens anyway if ___init__ is removed

Closes https://github.com/coala/corobo/issues/601

# Reviewers Checklist

### DIFF
--- a/plugins/coatils.py
+++ b/plugins/coatils.py
@@ -15,9 +15,6 @@ class Coatils(BotPlugin):
     Various coala related utilities, exposing the REST API, etc.
     """
 
-    def __init__(self, bot, name=None):
-        super().__init__(bot, name)
-
     @staticmethod
     def total_bears():
         bears = client.list.bears.get().json()


sladyn98:
The commit looks good to me.

jayvdb:
ack https://github.com/coala/corobo/pull/604/commits/b9315d14df029d59142921bc781965467635085a

jayvdb:
@gitmate-bot ff

jayvdb:
`__new__` & `__init__` combined are the *implementation* of the contructor.  You can read about it at https://en.wikipedia.org/wiki/Constructor_(object-oriented_programming) , especially https://en.wikipedia.org/wiki/Constructor_(object-oriented_programming)#Python

`Foo()` is *never* a constructor.  It is invoking a constructor.  It is rarely, and always incorrectly, called the constructor.  It should be called instantiation.

PR #613, opened 2018-08-09
 setup.cfg: Use template for enforcing 100% branch coverage

jayvdb:
# Reviewers Checklist


utils/mixin.py
```diff
@@ -20,6 +20,8 @@ def get_configuration_template(self):
             return default_config
         elif self.CONFIG_TEMPLATE:
             return self.CONFIG_TEMPLATE
+        else:  # pragma: no cover
```
RaiVaibhav:
```python
    def get_configuration_template(self):
        default_config = self._default_config
        if default_config:
            return default_config
        else self.CONFIG_TEMPLATE:
            return self.CONFIG_TEMPLATE
        return
```

**?**

plugins/searchdocs.py
```diff
@@ -30,7 +30,7 @@ def search(self, msg, arg):
             return (self.API_DOCS +
                     '/search.html?q=' +
                     '+'.join(re.split(r'\s+', search_string)))
-        elif doc_type.lower() == 'user':
+        else:
```
RaiVaibhav: No need for else IMO when this will be the last condition instead directly return

plugins/coatils.py
```diff
@@ -93,7 +93,7 @@ def contrib_stats(self, msg, match):
                 len(Coatils.all_langs())
             )
         # stats
-        elif stat_type == '' and entity == '':
+        else:
```
RaiVaibhav: https://github.com/coala/corobo/pull/613/files#r208848778

plugins/answer.py
```diff
@@ -20,6 +20,8 @@ def construct_link(text):
         elif 'documentation/' in text:
             text = text.split('documentation/')[-1]
             return 'https://docs.coala.io/en/latest/' + text
+        else:  # pragma: no cover
```
RaiVaibhav: https://github.com/coala/corobo/pull/613/files#r208848778

utils/mixin.py
jayvdb: `else self.CONFIG_TEMPLATE:` is syntax error ;-)

RaiVaibhav: oh okay :+1:

RaiVaibhav: these were a minor review, discard below comment directly if there will be some error :)

plugins/searchdocs.py
jayvdb: My own words thrown back at me ? ;)

plugins/coatils.py
jayvdb: This isnt as simple as the other one.  Each branch yields, and then goes to the end.

To remove this else branch, we would need to add returns to all of the other branches, which isnt very helpful.

PR #639, opened 2018-10-24
Added 'Zulip' as option for ROOMS_TO_JOIN

### LINKED ISSUE #635: On Zulip, ROOMS_TO_JOIN should contain valid Zulip rooms
Following on from https://github.com/coala/corobo/pull/634 , the rooms list needs to be updated on backend Zulip to contain only rooms which exist on the coala Zulip server , otherwise there are lots of errors on startup.

Note that 'maintainers' (same as gitter coala/coala/maintainers) and 'gci-mentors-2018' (no gitter equivalent) are two private rooms the bot also needs to join on Zulip.

why-does-ie-still-exist:
Added this code which if the `BACKEND` is run as `'Zulip'` ROOMS_TO_JOIN is updated accordingly.
I noticed there's not much code for Zulip(a cursory search shows no matches in the repo). So this will probably make someone's life easier down the line when they port the bot.
:sparkles: 

Third time's the charm :crossed_fingers: 

closes #635 
# Reviewers Checklist

### DIFF
--- a/config.py
+++ b/config.py
@@ -89,31 +89,46 @@ IGNORE_USERNAMES = os.environ.get("IGNORE_USERNAMES",
 
 DIVERT_TO_PRIVATE = ('help', )
 
-ROOMS_TO_JOIN = (
+ROOMS_TO_JOIN = [
     'coala',
-    'coala/offtopic',
-    'cobot-test',
-    'corobo',
-    'devops',
-    'community',
-    'coala/gsoc',
-    'coala/maintainers',
     'coala-bears',
-    'bearship',
-    'gci',
-    'cobot',
-    'performance',
-    'documentation',
-    'conferences',
-    'coala/workshops',
-    'coala/artwork-corner',
-    'freelancers',
-    'editor-plugins',
+    'corobo',
     'depman',
     'ast',
-    'aspects',
-    'community'
-)
+    'gci',
+]
+
+if BACKEND == 'Gitter':
+    ROOMS_TO_JOIN += [
+        'aspects',
+        'bearship',
+        'coala',
+        'coala/artwork-corner',
+        'coala/gsoc',
+        'coala/maintainers',
+        'coala/offtopic',
+        'coala/workshops',
+        'cobot',
+        'cobot-test',
+        'community',
+        'community',
+        'conferences',
+        'devops',
+        'documentation',
+        'editor-plugins',
+        'freelancers',
+        'performance',
+    ]
+elif BACKEND == 'Zulip':
+    ROOMS_TO_JOIN += [
+        'maintainers',
+        'gci-mentors-2018',
+        'gitmate',
+        'gsoc',
+        'moban',
+        'test',
+        'zulip',
+    ]
 
 if BACKEND == 'Gitter':
     ROOMS_TO_JOIN = ['coala/' + item for item in ROOMS_TO_JOIN]


why-does-ie-still-exist:
@jayvdb Sorry to bother you...Why do I have pull request issues?
image

config.py
```diff
@@ -89,31 +89,47 @@
...
+
+if BACKEND == 'Zulip':
+    ROOMS_TO_JOIN = (
+        'announce',
+        'coala',
+        'coala/maintainers',
+        'coala-bears',
+        'corobo',
+        'depman',
+        'gci-mentors-2018'
```
jayvdb: There are more rooms in Zulip, even rooms which are the same as Gitter, e.g. `gci` and `ast`-- look for them all.

config.py
```diff
@@ -89,31 +89,47 @@
...
-    'coala/artwork-corner',
-    'freelancers',
-    'editor-plugins',
-    'depman',
-    'ast',
-    'aspects',
-    'community'
-)
+if BACKEND == 'Gitter':
+    ROOMS_TO_JOIN = (
```
jayvdb: and then

```py
if BACKEND == 'Gitter':
    ROOMS_TO_JOIN += [
        'cobot-test',
        ...,
    ]
elif BACKEND == 'Zulip':
    ROOMS_TO_JOIN += [
        'announce',
        'gci-mentors-2018',
        # any others?
    ]

File: config.py [hunk 3]
```diff
@@ -89,31 +89,47 @@
 
 DIVERT_TO_PRIVATE = ('help', )
 
-ROOMS_TO_JOIN = (
```
jayvdb: It would be better to have all rooms which are common stay in the variable at the top here, ..

but change it from a tuple `(..)` to a list `[ ... ]`

so this would be

```py
ROOMS_TO_JOIN = [
```

config.py
why-does-ie-still-exist: I can't see any other rooms.

config.py
why-does-ie-still-exist: Got it.

config.py
why-does-ie-still-exist: Ok.

config.py
```diff
@@ -89,31 +89,46 @@
...
+        'community',
+        'community',
+        'conferences',
+        'devops',
+        'documentation',
+        'editor-plugins',
+        'freelancers',
+        'gci',
+        'performance'
+    )
```
Vamshi99: Can you move this https://github.com/coala/corobo/blob/master/config.py#L119 here?

jayvdb: This `)` is a syntax error.  A list ends with a `]`

sks444: Also, don't forget to add a trailing comma.

config.py
```diff
@@ -89,31 +89,46 @@
...
+elif BACKEND == 'Zulip':
+    ROOMS_TO_JOIN += [
+        'announce',
+        'maintainers',
+        'gci-mentors-2018'
+        'gitmate',
+        'offtopic',
+        'test',
+        'zulip'
```
sks444: here too, trailing comma.

config.py
jayvdb: Did you look at all streams, even the ones you are not subscribed too ?

config.py
```diff
@@ -89,31 +89,46 @@
...
-    'ast',
-    'aspects',
-    'community'
-)
+]
+
+if BACKEND == 'Gitter':
+    ROOMS_TO_JOIN += [
+        'aspects',
+        'ast',
```
jayvdb: As mentioned previously, `ast` is a room in both gitter and zulip

config.py
```diff
@@ -89,31 +89,46 @@
...
+        'bearship',
+        'coala',
```
jayvdb: `coala` is a room in both gitter and zulip .  doesnt need to be repeated here, as it is the list above

config.py
```diff
@@ -89,31 +89,46 @@
...
+        'cobot',
+        'cobot-test',
+        'community',
+        'community',
+        'conferences',
+        'devops',
+        'documentation',
+        'editor-plugins',
+        'freelancers',
+        'gci',
```
jayvdb: 'gci' room is common - it should be in the first list

config.py
why-does-ie-still-exist: I found all the streams! I had to click search. It should soon be done.

why-does-ie-still-exist:
@jayvdb, other than the two chats with the hindi and sindhi script, (which for some reason notepad gets very angry about) I think I've got it :-)

why-does-ie-still-exist:
gotta squash those commits

config.py
```diff
@@ -89,31 +89,49 @@
...
+        'documentation',
+        'editor-plugins',
+        'freelancers',
+        'performance',
+    ]
+elif BACKEND == 'Zulip':
+    ROOMS_TO_JOIN += [
+        'announce',
+        'maintainers',
+        'gci-mentors-2018'
```
jayvdb: we need a trailing comma here

config.py
```diff
@@ -89,31 +89,49 @@
...
+        'gci bahasa'
```
jayvdb: and trailing comma here

config.py
```diff
@@ -89,31 +89,49 @@
...
+        'gitmate',
+        'gsoc',
+        'moban',
+        'offtopic',
```
jayvdb: ah , this is one room which we dont want the bot in.

jayvdb:
ack https://github.com/coala/corobo/pull/639/commits/302bce4bbe459be5eba12b48dbe7ddcb816ed916

jayvdb:
@gitmate-bot ff

Issue #632, opened 2018-10-24
labhub: Add error message if nick is empty


jayvdb:
Currently labhub uses the `nick` as the username, because this is how gitter works.

But not all backends provide a `nick`.  See https://github.com/coala/corobo/issues/255

All labhub commands which need `nick` should provide an error message

jayvdb:
This includes at least `assign` and `unassign`.

PR #642, opened 2018-11-02
labhub.py: Add error message when nick is empty

abhishalya:
All of the backends do not have `nick` attribute.
Hence an error will be shown whenever nick is empty.

Closes https://github.com/coala/corobo/issues/632

# Reviewers Checklist

- [x] Appropriate logging is done.
- [x] Appropriate error responses.
- [x] Handle every possible exception.
- [x] Make sure there is a docstring in the command functions. Hint: Lookout for
- [x] See that 100% coverage is there.
- [x] See to it that mocking is not done where it is not necessary.

### DIFF
--- a/plugins/labhub.py
+++ b/plugins/labhub.py
@@ -122,6 +122,9 @@ class LabHub(DefaultConfigMixin, BotPlugin):
         """
         invitee = match.group(1)
         inviter = msg.frm.nick
+        if not inviter:
+            yield 'ERROR: The above command cannot be operated without nick.'
+            return
 
         team = 'newcomers' if match.group(2) is None else match.group(2)
         team = team.lower()
@@ -196,6 +199,9 @@ class LabHub(DefaultConfigMixin, BotPlugin):
     def create_issue_cmd(self, msg, match):
         """Create issues on GitHub and GitLab repositories."""  # Ignore QuotesBear, LineLengthBear, PyCodeStyleBear
         user = msg.frm.nick
+        if not user:
+            yield 'ERROR: The above command cannot be operated without nick.'
+            return
         repo_name = match.group(1)
         iss_title = match.group(2)
         iss_description = match.group(3) if match.group(3) is not None else ''
@@ -234,6 +240,9 @@ class LabHub(DefaultConfigMixin, BotPlugin):
         issue_number = match.group(4)
 
         user = msg.frm.nick
+        if not user:
+            yield 'ERROR: The above command cannot be operated without nick.'
+            return
 
         try:
             assert org == self.GH_ORG_NAME or org == self.GL_ORG_NAME
@@ -317,6 +326,9 @@ class LabHub(DefaultConfigMixin, BotPlugin):
         iss_number = match.group(4)
 
         user = msg.frm.nick
+        if not user:
+            yield 'ERROR: The above command cannot be operated without nick.'
+            return
 
         try:
             assert org == self.GH_ORG_NAME or org == self.GL_ORG_NAME
--- a/tests/labhub_test.py
+++ b/tests/labhub_test.py
@@ -33,6 +33,7 @@ class TestLabHub(LabHubTestCase):
             'GH_ORG_NAME': 'coala',
             'GL_ORG_NAME': 'coala',
         }
+        self.bot.sender._nick = 'batman'
         self.labhub = self.load_plugin('LabHub', self.global_mocks, configs)
 
     def test_invite_cmd(self):
@@ -102,6 +103,11 @@ class TestLabHub(LabHubTestCase):
         testbot.assertCommand('!invite meetto newcomers',
                               'Command "invite" / "invite meetto" not found.')
 
+        self.bot.sender._nick = None
+        testbot.assertCommand(
+            '!invite meet to newcomers',
+            'ERROR: The above command cannot be operated without nick.')
+
         # not a member of org
         mock_team_newcomers.is_member.return_value = False
         mock_team_developers.is_member.return_value = False
@@ -156,7 +162,7 @@ class TestLabHub(LabHubTestCase):
                 textwrap.dedent('''\
                     first line of body
                     second line of body
-                    Opened by @None at [text]()''')
+                    Opened by @batman at [text]()''')
         )
 
         testbot_public.assertCommand(
@@ -171,12 +177,17 @@ class TestLabHub(LabHubTestCase):
                 'another title',
                 textwrap.dedent('''\
                     body
-                    Opened by @None at [text]()''')
+                    Opened by @batman at [text]()''')
         )
 
         testbot_public.assertCommand(
             '!new issue coala title',
             'repository that does not exist')
+
+        self.bot.sender._nick = None
+        testbot_public.assertCommand(
+            '!new issue coala title',
+            'ERROR: The above command cannot be operated without nick.')
 
         # not a member of org
         self.mock_team.is_member.return_value = False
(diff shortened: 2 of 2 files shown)


abhishalya:
I apologize for not checking the code locally. WIll not push until checked :)

plugins/labhub.py
```diff
@@ -92,6 +92,10 @@ def team_mapping(self):
     @cmdfilter
     def members_only(self, msg, cmd, args, dry_run):
         user = msg.frm.nick
+        if not user:
+            self.send(msg.frm, 'ERROR: The above command cannot be '
+                      'operated using nick')
```
jayvdb: sentence logic error here

using -> without

Also it is not good style to have the start of the strings in different columns.
Instead, if it doesnt all fit on one line, start the string on a new line, like:
```py
            self.send(msg.frm,
                      'ERROR: The above command cannot be '
￼                      'operated using nick')
￼```

File: plugins/labhub.py [hunk 2]
```diff
@@ -122,6 +126,10 @@ def invite_cmd(self, msg, match):
         """
         invitee = match.group(1)
         inviter = msg.frm.nick
+        if not inviter:
+            self.send(msg.frm, 'ERROR: The above command cannot be '
```
jayvdb: errors should be yielded

abhishalya: Got it :+1:

plugins/labhub.py
abhishalya: Oh, okay. I'll make the changes

abhishalya: Similar line was found at https://github.com/coala/corobo/blob/master/plugins/labhub.py#L101.

plugins/labhub.py
nvzard: @abhishalya, since `members_only` is a command filter and it is called before the desired bot command method is called. So you don't need to check for `nick` in every command method.

plugins/labhub.py
nvzard: :point_up: That was me there :upside_down_face:

abhishalya: @nvzard Why thumbs down? You could've improved there :wink:

plugins/labhub.py
abhishalya: @nvzard Yeah, I knew that. I suggested that on zulip, but since no one responded I appended the same code to all. Will do this change :+1:

tests/labhub_test.py
```diff
@@ -53,6 +53,14 @@ def test_invite_cmd(self):
...
         self.assertEqual(self.labhub.TEAMS, self.teams)
 
+        testbot.assertCommand(
+            '!invite meet to newcomers',
+            'ERROR: The above command cannot '
+            'be operated without nick')
+        mock_msg = create_autospec(Message)
+        mock_msg.frm.nick = PropertyMock()
+        mock_msg.frm.nick = 'batman'
+
```
nvzard: CI is failing because you are testing it wrong. We need to pass the mocked message into the bot but that is not possible at the moment while testing.
I suggest you create a class method to check if `nick` is present or not and call that for the required bot commands so that it'd be easier to test. Like this: https://github.com/coala/corobo/blob/b9d569d1461d5ffe11b8f04766221b71b8e490f9/plugins/labhub.py#L112
 Write the tests the same way I did for 'is_room_member' method.

tests/labhub_test.py
```diff
@@ -53,6 +53,14 @@ def test_invite_cmd(self):
 
         self.assertEqual(self.labhub.TEAMS, self.teams)
 
+        testbot.assertCommand(
```
jayvdb: as I suggested in DM, create a new test method, dont edit the existing one.

That way, the existing tests can not possibly break unless your code is wrong.

abhishalya: They will break, since I'm introducing a new error on the functions present in the lanhub.py file. I'll have to either create a mock object which will ensure that the validate_nick function will return true or I can remove the error itself and just introduce the function and its test, but that won't solve the issue.

abhishalya: I'm pushing the code, so that you can review better.

abhishalya:
Btw I got an InvalidLinkBear is complaining:

```bash
plugins/ship_it.py
|  14| ········'http://i.imgur.com/DPVM1.png',
|    | [NORMAL] InvalidLinkBear:
|    | Broken link - unable to connect to http://i.imgur.com/DPVM1.png (HTTP Error: 404)

```

plugins/labhub.py
```diff
@@ -13,6 +13,12 @@
 from utils.mixin import DefaultConfigMixin
 
 
+def validate_nick(msg):
```
meetmangukiya: I fail to understand why you are doing this. This function merely returns `bool(msg.frm.nick)`. You can just use `if msg.frm.nick:` instead of `if validate_nick(msg)`.. This function does no good.

tests/labhub_test.py
```diff
@@ -59,6 +60,9 @@ def test_invite_cmd(self):
 
         mock_dict['is_room_member'].return_value = True
 
+        mock_nick = create_autospec(Message)
+        mock_nick.frm.nick.return_value = True
+
```
meetmangukiya: `mock_nick` isn't used anywhere in `test_invite_cmd`... unused code.

tests/labhub_test.py
```diff
@@ -117,6 +121,12 @@ def test_is_room_member(self):
         msg.frm.room.occupants = ['batman', 'superman']
         self.assertTrue(LabHub.is_room_member('batman', msg))
 
+    def test_validate_nick(self):
```
meetmangukiya: Since I think the function itself is not needed, so this test function is irrelevant as well..

tests/labhub_test.py
```diff
@@ -142,6 +152,9 @@ def test_create_issue_cmd(self):
         self.mock_team.is_member.return_value = True
         testbot_public = self
 
+        mock_nick = create_autospec(Message)
+        mock_nick.frm.nick.return_value = True
+
```
meetmangukiya: Creating these `mock_nick`s aren't doing any good, because you are not passing the mocked `Message` object anywhere.

tests/labhub_test.py
```diff
@@ -241,6 +256,8 @@ def test_assign_cmd(self):
         mock_maint_team = create_autospec(github3.orgs.Team)
         mock_dev_team.is_member.return_value = False
         mock_maint_team.is_member.return_value = False
+        mock_nick = create_autospec(Message)
```
jayvdb: please undo all of your changes to other test methods.

First get one new test method to work, then you can duplicate it.

abhishalya: Doing this right now.

tests/labhub_test.py
abhishalya: Then how do I do this? The tests system is not comfortable with changes.

plugins/labhub.py
```diff
@@ -122,6 +128,8 @@ def invite_cmd(self, msg, match):
         """
         invitee = match.group(1)
         inviter = msg.frm.nick
+        if not validate_nick(msg):
+            yield 'ERROR: The above command cannot be operated without nick.'
```
nvzard: @abhishalya  You need to ` return` after yielding. Check this.

tests/labhub_test.py
```diff
@@ -12,6 +12,7 @@
 
 import plugins.labhub
 from plugins.labhub import LabHub
+from plugins.labhub import validate_nick
```
nvzard: we can also do `from plugins.labhub import LabHub, validate_nick`

jayvdb:
ack https://github.com/coala/corobo/pull/642/commits/4c613935607a16a6fd7530ee287f5089326e91bc

jayvdb:
@gitmate-bot ff

Issue #347, opened 2017-11-22
GitLab issues created on Gitter should link to github.com/<username>


jayvdb:
Since the issues are opened via gitter, the username of github is used, and it is not always safe to assume same username for gitlab as well

meetmangukiya:
Ideally a new function should be created in https://github.com/coala/corobo/blob/master/utils/backends.py which make relevant guesses on the basis of backend. So, if it is `gitter` map it to `github`, if it is `mattermost` most likely option is `gitlab`, etc.

sladyn98:
Before I get this issue assigned to me. I would like to propose the following solution
1) If the backend is gitter check if it is ```gitlab``` or ```github``` and construct the username as follows:
https://github.com/username similar for gitlab
2) If the backend is mattermost ```gitlab``` would be sufficient i guess.
3) Return the profile link.
Additional notes:
A proper regex expression for extracting the underscore ```^(.*)(_.*?)$```.
If all clear then kindly assign me this issue.Thanks a lot

PR #645, opened 2018-12-20
backends.py: Add profile_link function

sladyn98:
This function is created to map github and gitlab usernames.
Ensures correct profile links are mapped for both websites.

Fixes https://github.com/coala/corobo/issues/347

# Reviewers Checklist

### DIFF
--- a/plugins/labhub.py
+++ b/plugins/labhub.py
@@ -206,7 +206,7 @@ class LabHub(DefaultConfigMixin, BotPlugin):
         iss_title = match.group(2)
         iss_description = match.group(3) if match.group(3) is not None else ''
         extra_msg = '\nOpened by @{username} at [{backend}]({msg_link})'.format(
-            username=user,
+            username=self.profile_link(msg),
             backend=self.bot_config.BACKEND,
             msg_link=message_link(self, msg)
         )
@@ -221,6 +221,17 @@ class LabHub(DefaultConfigMixin, BotPlugin):
             ).render(
                 target=user,
             )
+
+    @staticmethod
+    def profile_link(msg):
+        gitter_username = msg.frm.nick
+        if '_gitlab' in gitter_username:
+            pattern = '^(.*)(_.*?)$'
+            username = re.match(
+                pattern, gitter_username.split('@')[-1]).group(1)
+        else:
+            username = gitter_username.split('@')[-1]
+        return 'https://github.com/{user}'.format(user=username)
 
     @staticmethod
     def is_newcomer_issue(iss):
--- a/tests/labhub_test.py
+++ b/tests/labhub_test.py
@@ -162,7 +162,7 @@ class TestLabHub(LabHubTestCase):
                 textwrap.dedent('''\
                     first line of body
                     second line of body
-                    Opened by @batman at [text]()''')
+                    Opened by @https://github.com/batman at [text]()''')
         )
 
         testbot_public.assertCommand(
@@ -177,7 +177,7 @@ class TestLabHub(LabHubTestCase):
                 'another title',
                 textwrap.dedent('''\
                     body
-                    Opened by @batman at [text]()''')
+                    Opened by @https://github.com/batman at [text]()''')
         )
 
         testbot_public.assertCommand(
@@ -206,6 +206,16 @@ class TestLabHub(LabHubTestCase):
         self.assertTrue(LabHub.is_newcomer_issue(mock_iss))
         mock_iss.labels = ('difficulty/medium',)
         self.assertFalse(LabHub.is_newcomer_issue(mock_iss))
+
+    def test_profile_link(self):
+        msg = create_autospec(Message)
+        msg.frm.nick = PropertyMock()
+        msg.frm.nick = ('@sladyn98')
+        self.assertEqual(LabHub.profile_link(msg),
+                         'https://github.com/sladyn98')
+        msg.frm.nick = ('@sladyn98_gitlab')
+        self.assertEqual(LabHub.profile_link(msg),
+                         'https://gitlab.com/sladyn98')
 
     def test_unassign_cmd(self):
         self.inject_mocks('LabHub', {'REPOS': {'example': self.mock_repo}})


abhishalya:
@sladyn98 The changes you made I think are only valid for gitter as backend. You need to find the ways of mapping usernames for other backends too. Also, you need to use this function while creating the issue.

sladyn98:
@abhishalya but I think the issue description mentions issues created only via gitter right?In terms of  other backends we would by default have to just make a guess for the username because we won't be absolutely sure that they are the original gitlab/github usernames.

sladyn98:
If the issue demands In terms of mattermost we could return gitlab. For slack and other backends,I think github would be the best guess. Thoughts?

abhishalya:
@sladyn98 Well then you should not make changes to the backend.py file. Instead I feel the function should be added to the labhub because that is where the issue is being created.

sladyn98:
Yeah @abhishalya I totally agree but that wouldn't change much because we would be doing the same things there as well.

abhishalya:
I think the functions in the backend.py file are the ones where the output is dependent on the backend used by the corobo. Here since we can only form solution for gitter backend we should not add that function to the backend file. Also, currently the function does nothing, we are not using it. You'll need to form the entire solution and then test it too.

sladyn98:
Yeah so for now I think we can add the gitter backend functionality to
labhub and test out the functionality,i don't see the point of other
backends because we would return only the github profile for that username
based on a guess. The other backend we can add is mattermost so I think
I'll add a function to labhub and integrate it with create issues and
create a PR.

sladyn98:
We can then test out this  functionality,because most of the issues would be created via gitter so this solution should be more than enough

sladyn98:
@abhishalya yeah buddy i wil write the tests is everthing good till now.

abhishalya:
@sladyn98 `git grep message_link`

sladyn98:
@abhishalya I did a grep search but no such tests popped up.

tests/labhub_test.py
```diff
@@ -207,6 +207,13 @@ def test_is_newcomer_issue(self):
         mock_iss.labels = ('difficulty/medium',)
         self.assertFalse(LabHub.is_newcomer_issue(mock_iss))
 
+    def test_profile_link(self):
+        link = profile_link("sladyn98")
```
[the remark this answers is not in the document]
sladyn98: @abhishalya you mean i should just create a mock object for the string and assert true on that mock object ?

utils/backends.py
```diff
@@ -1,7 +1,6 @@
 """
 Handle backend specific implementations.
 """
-
```
abhishalya: Don't change this file, it is not a part of your issue.

abhishalya:
Also please run `coala` before pushing, so as to avoid bunch of comments from the gitmate-bot.

sladyn98:
> Also please run `coala` before pushing, so as to avoid bunch of comments from the gitmate-bot.

Yeah Im sorry I will run that.

sladyn98:
@abhishalya now the tests are passing and the coverage has also gone up.

plugins/labhub.py
```diff
@@ -222,6 +222,18 @@ def create_issue_cmd(self, msg, match):
                 target=user,
             )
 
+    @staticmethod
+    def profile_link(self, msg):
```
nvzard: Why does a `staticmethod` needs `self` object?

tests/labhub_test.py
```diff
@@ -207,6 +207,13 @@ def test_is_newcomer_issue(self):
         mock_iss.labels = ('difficulty/medium',)
         self.assertFalse(LabHub.is_newcomer_issue(mock_iss))
 
+    def test_profile_link(self):
+        msg = create_autospec(Message)
+        msg.frm.nick = PropertyMock()
+        msg.frm.nick = ('@sladyn98')
+        self.assertEqual(LabHub.profile_link(self, msg),
+                         'https://github.com/sladyn98')
+
```
nvzard: Add a test for the a gitlab user

plugins/labhub.py
sladyn98: Oh I might have committed the old file :(
Fixed

tests/labhub_test.py
sladyn98: Added :P

tests/labhub_test.py
```diff
@@ -207,6 +207,16 @@ def test_is_newcomer_issue(self):
         mock_iss.labels = ('difficulty/medium',)
         self.assertFalse(LabHub.is_newcomer_issue(mock_iss))
 
+    def test_profile_link(self):
+        msg = create_autospec(Message)
+        msg.frm.nick = PropertyMock()
+        msg.frm.nick = ('@sladyn98')
+        self.assertEqual(LabHub.profile_link(self, msg),
```
Vamshi99: As you're using a static method, don't use self here. Same in the GitLab case

plugins/labhub.py
```diff
@@ -222,6 +222,18 @@ def create_issue_cmd(self, msg, match):
                 target=user,
             )
 
+    @staticmethod
+    def profile_link(msg):
+        gitter_username = msg.frm.nick
+        pattern = '^(.*)(_.*?)$'
```
Makman2: you should "declare" where it's used

plugins/labhub.py
```diff
@@ -222,6 +222,18 @@ def create_issue_cmd(self, msg, match):
...
+        if '_gitlab' in gitter_username:
+            username = re.match(
+                pattern, gitter_username.split('@')[-1]).group(1)
+            return 'https://gitlab.com/{user}'.format(user=username)
+        else:
+            username = gitter_username.split('@')[-1]
+            return 'https://github.com/{user}'.format(user=username)
```
Makman2: duplicated logic here.

plugins/labhub.py
```diff
@@ -222,6 +222,18 @@ def create_issue_cmd(self, msg, match):
...
             )
 
+    @staticmethod
+    def profile_link(msg):
+        gitter_username = msg.frm.nick
+        pattern = '^(.*)(_.*?)$'
+        if '_gitlab' in gitter_username:
+            username = re.match(
+                pattern, gitter_username.split('@')[-1]).group(1)
+            return 'https://gitlab.com/{user}'.format(user=username)
```
Makman2: Btw you sure it reads so well providing the full url? I would maybe do something like `@<username> (GitLab)` or similar

Makman2: Ah okay now I see... It's otherwise linking to the user. Is it possible to override the link via `[]()` syntax?

plugins/labhub.py
sladyn98: I don't quite seem to understand what do you want me to declare :)

Makman2: You assign to pattern although it's just use in one of the if branches. You should do the assignment there^^

plugins/labhub.py
abhishalya: Yeah, right now it would look something like:

Opened by @https://gitlab.com/abhishalya at Gitter.

A simple change to https://github.com/coala/corobo/blob/master/plugins/labhub.py#L208 would fix this.

plugins/labhub.py
```diff
@@ -222,6 +222,18 @@ def create_issue_cmd(self, msg, match):
...
+        else:
```
abhishalya: Isin't there login through twitter option too on Gitter? You need to get that too imo.

sladyn98: @abhishalya IMO not a lot of users use twitter so that could be left out for now.

plugins/labhub.py
sladyn98: Done :+1:

plugins/labhub.py
sladyn98: Changed :+1:

plugins/labhub.py
sladyn98: Yeah we could override it.We could even do something like 
```hyperlink_format = '<a href="{link}">{text}</a>```
```hyperlink_format.format(link='https://gitlab.com/sladyn98', text='@sladyn98')```

abhishalya: No, we return a markdown string to gitter.

plugins/labhub.py
abhishalya: No, again, that should be a part of this issue itself.

sladyn98:
@Makman2 Kindly review this PR.

PR #650, opened 2019-02-27
Fix w605

sladyn98:
# Reviewers Checklist

### DIFF
--- a/.ci/semaphore.tests.sh
+++ b/.ci/semaphore.tests.sh
@@ -1,6 +1,6 @@
 set -e -x
 
-pip install coala-bears
+pip install git+https://github.com/coala/coala-bears.git@master
 pip install -r requirements.txt
 pip install -r test-requirements.txt
 sed -i.bak '/bears = GitCommitBear/d' .coafile
--- a/plugins/pitchfork.py
+++ b/plugins/pitchfork.py
@@ -21,13 +21,13 @@ class Pitchfork(BotPlugin):
             user = match.group(1)
             place = match.group(2) if match.group(2) else 'offtopic'
             return textwrap.dedent((
-                string.Template("""
+                string.Template(r"""
                     @$user, you are being pitchforked down to $place
                     ```
                                                           .+====----->
-                                                           \('
+                                                           \\('
                     =====================================<%{%{%{>>+===---> $user
-                                                           /(,
+                                                           //(,
                                                           .+====----->
                     ```
                 """).substitute(user=user,


plugins/pitchfork.py
```diff
@@ -25,7 +25,7 @@ def pitchfork(self, msg, arg):
                     @$user, you are being pitchforked down to $place
                     ```
                                                           .+====----->
-                                                           \('
+                                                           \\('
                     =====================================<%{%{%{>>+===---> $user
                                                            /(,
```
nvzard: Make this like `//(,` just to make the trident symmetric?

sladyn98: Yeah this could be done as well , but is it really required :P

nvzard: Yes.

Makman2: Actually it has to be just here. The double escape will in the chat print as a single one, so making below a double `//` would then look weird in the output. You could probably just prepend `r` to the string to make it a raw string so it keeps symmetric in code and in output.

plugins/pitchfork.py
```diff
@@ -25,9 +25,9 @@ def pitchfork(self, msg, arg):
                     @$user, you are being pitchforked down to $place
                     ```
                                                           .+====----->
-                                                           \('
+                                                           \\('
                     =====================================<%{%{%{>>+===---> $user
-                                                           /(,
+                                                           //(,
```
Makman2: See previous comment: This makes output "asymmetric".

.ci/semaphore.tests.sh
```diff
@@ -1,9 +1,13 @@
...
+cp requirements.txt requirements.orig
+printf '%s\n%s\n%s\n%s\n' \
+    "git+https://github.com/coala/coala" \
+    "git+https://github.com/coala/coala-bears" \
+    "$(cat test-requirements.txt)" \
+    "$(cat requirements.txt)" \
+    > requirements.txt
 pip install -r requirements.txt
-pip install -r test-requirements.txt
-sed -i.bak '/bears = GitCommitBear/d' .coafile
```
Makman2: Can't we just fix to the latest dev-version inside `test-requirements.txt`? Then we don't introduce silent breakages by updates in the coala and coala-bears breakages without an explicit upgrade here^^

plugins/pitchfork.py
sladyn98: Yeah your probably right, I will prepend r to it so that it makes the code symmetric.

.ci/semaphore.tests.sh
sladyn98: Sound like a plan :)
Should I ammend this commit?

nvzard: Yes, @sladyn98 amend this commit.

sladyn98: I am not very familiar with how the test-requirements have been written. But correct me if I am wrong these two lines would be added to `test - requirements.txt`
`git+https://github.com/coala/coala`
`git+https://github.com/coala/coala-bears`
And the above sh file would now contain the following 
```pip install -r requirements.txt```
```pip install -r test-requirements.txt```
```sed -i.bak '/bears = GitCommitBear/d' .coafile```

**Comment by abhishalya** [comment]
@sladyn98 Please fix this PR.

**Comment by sladyn98** [comment]
Yeah @abhishalya can you please verify the above comment, because I am not very familiar with coorbo's internal requirements :)

File: .ci/semaphore.tests.sh (hunk 3 above)
Comment (abhishalya) [reply .ci/semaphore.tests.sh]: Yeah, that should work @sladyn98

**Comment by nvzard** [comment]
@sladyn98 what you need to do is install the latest commit from coala/coala-bears which works without breaking our CI. By doing this we are fixing the dev version of coala-bears that we use, so that our CI won't break if some new addition in coala-bears causes breakage.

**Comment by sladyn98** [comment]
@nvzard 
@nvzard
Yeah sure but which branch or release do you want me to add
If you want to add a hash we would have to do something like this
```
git clone [URLTORepository]
git checkout  [commithash]
git clone --branch <tag_name> <repo_url>
```
If you want to add a tag we could do something like this

**Comment by abhishalya** [comment]
@sladyn98 You forgot to edit the .travis.yml file, you don't need to move coala, coala-bears requirements manually to the requirememts.txt file. I think it should be fine to use the latest coala, coala-bears since they work fine right now.

**Comment by sladyn98** [comment]
@abhishalya IMO we could add a commit or a tag to the links which would make much more sense

**Comment by abhishalya** [comment]
@sladyn98 No, we don't follow that. For every repo, we need the latest coala, coala-bears so that any fixes in those would reflect for all. And currently both of them work as expected, so we don't need to switch to a particular commit.

File: plugins/pitchfork.py [hunk 4]
```diff
@@ -21,11 +21,11 @@ def pitchfork(self, msg, arg):
...
             return textwrap.dedent((
-                string.Template("""
+                string.Template(r"""
                     @$user, you are being pitchforked down to $place
                     ```
                                                           .+====----->
-                                                           \('
+                                                           \\('
                     =====================================<%{%{%{>>+===---> $user
                                                            /(,
```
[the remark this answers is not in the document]
sladyn98: Yeah agreed :) I will get that fixed asap.

sladyn98:
> Add travis.yml changes.
> 
> This PR needs to be fixed as soon as possible.

Yep working on it :)

Dockerfile
```diff
@@ -4,10 +4,8 @@ ENV BOT_ROOT=/opt/errbot
 
 ADD requirements.txt $BOT_ROOT/requirements.txt
 
-RUN apk add --no-cache libffi openssl git \
+RUN apk add --no-cache libffi openssl git gcc libxml2-dev libxslt-dev libc-dev\
```
bkhanale: I think its better the previous way (as currently implemented in travis.yml). So no modifications required here.

Dockerfile
```diff
@@ -4,10 +4,8 @@ ENV BOT_ROOT=/opt/errbot
...
     && apk add --no-cache --virtual .build-deps \
-           gcc \
```
bkhanale: Undo as per previous comment.

test-requirements.txt
```diff
@@ -6,6 +6,8 @@ colorama<0.4
 # coverage 4.4.2 broke compatibility with coverage-config-reload-plugin
 # and thus broke https://github.com/jayvdb/coverage_env_plugin .
 # See https://github.com/jayvdb/coverage_config_reload_plugin/issues/1
+git+https://github.com/coala/coala@master
+git+https://github.com/coala/coala-bears@master
```
bkhanale: Undo here as well.

sladyn98: We need to pull the latest coala bears repo from master so we are in dicussions to see whether it works.

bkhanale: >We need to pull the latest coala bears repo from master so we are in dicussions to see whether it works.

Yes, I know that very well. To implement that we shouldn't do modifications to the test-requirements since we'll have to do a lot of changes to Dockerfile which would eventually increase the size of the image. Since we are only using it (coala-bears) on travis we should install them separately (this is what we've already done on travis, we just need to follow the same for semaphore).

bkhanale:
>We need to pull the latest coala bears repo from master so we are in dicussions to see whether it works.

Yes, I know that very well. To implement that we shouldn't do modifications to the test-requirements since we'll have to do a lot of changes to Dockerfile which would eventually increase the size of the image. Since we are only using it (coala-bears) on travis we should install them separately (this is what we've already done on travis, we just need to follow the same for semaphore).

abhishalya:
Yeah, @bkhanale is right here. Sorry if I wasn't very clear earlier. This should help you fix this:

1. No changes to travis.yml file.
2. The changes made to `pitchfork.py` are good, and should be kept.
3. No changes to Dockerfile.
4. Change to .ci/semaphore.tests.sh should like travis.yml where we were explicitly installing coala and coala-bears through their github source.
5. No changes to test-requirements.txt.

abhishalya:
@sladyn98 Let me know if you would like to work on this, since we need the CI green to be seen at one spot I have to cherry pick your commits into mine. You could keep one commit here, and the other one I could add myself. That would require you to squash current two commits into one (the pitchfork one) and remove the changes of the second commit (the coala one). After which I could pick your commit and then push the other commit into my PR after which we should be able to see the CI green on that PR.

sladyn98:
@abhishalya @bkhanale I agree with both of you, but I just spoke to @Makman2 on gitter and he suggested that adding the files to test-requirements would prevent breakages, and the increased size of the dockerfile would not make much of a difference if the build is green again.Thanks for the suggesstions.

abhishalya:
@sladyn98 It's redundant since corobo doesn't require coala. We are only using it to keep the source clean, creating a docker image with coala is completely unnecessary IMO. I had spoken about this with Makman2 a while back, he had agreed with this.

abhishalya:
>adding the files to test-requirements would prevent breakages

How? We are still doing the same thing.

>the increased size of the dockerfile would not make much of a difference

Not the Dockerfile, but the docker image. Its size would be increased and even if size isin't a problem, corobo users wouldn't need coala in docker since either they would already have it installed or they wouldn't need one (if corobo is being used outside coala).

sladyn98:
@abhishalya Agreed with you completely :)

.ci/semaphore.tests.sh
```diff
@@ -1,6 +1,6 @@
 set -e -x
 
-pip install coala-bears
+pip install git+https://github.com/coala/coala-bears.git@master
```
abhishalya: Where is coala? :/

abhishalya:
@sladyn98 Please don't make random changes and invoke unnecessary builds.

.ci/semaphore.tests.sh
Makman2: It's a dependency of coala-bears. That way we let coala-bears manage its compatible coala version bears are written for.

Makman2:
That change was requested by me.

Makman2:
ack 707a43e 1806316

abhishalya:
@Makman2  Why though?

.ci/semaphore.tests.sh
abhishalya: Ah, okay. :+1:

sladyn98:
Done :tada:

Makman2:
@gitmate-bot ff

abhishalya:
@Makman2 This is not installing latest coala and would rely on `coala-0.11.0`.
As seen from the [semaphore CI log](https://semaphoreci.com/coala/corobo/branches/master/builds/194)
```
Collecting coala (from coala-bears==0.12.0.dev99999999999999)
  Using cached https://files.pythonhosted.org/packages/cc/74/8b822224ec8b35c4cace3f5768b9573ad33c1a1152f75a61b6c1cb416f78/coala-0.11.0-py34.py35.py36-none-any.whl
```
This could potentially be harmful, different versions of coala and coala-bears could break many things.

Makman2:
If they don't break right now I wouldn't invest more time into it. When the versions get incompatible, the right way would be to specify a newer version in coala-bears's requirements :+1:

PR #651, opened 2019-03-09
plugins:Add plugin for saying no to 'sir'

AkshJain99:
# Reviewers Checklist


plugins/no_sir.py
```diff
@@ -0,0 +1,18 @@
...
+from errbot import BotPlugin, re_botcmd
+
+
+class no_sir(BotPlugin):
+  """
+  Do not use sir
+  """
+  def callback_message(self, msg):
+    emots = [':D']
+    if match_sir:
```
sladyn98: @AkshJain99 What are you trying to do here? You instead should use a regex expression because the word `sir` could be present anywhere in the message.

plugins/no_sir.py
```diff
@@ -0,0 +1,18 @@
...
+      self.send(
+        msg.frm,
+        '@{}, Do not use sir in your conversation. {}'.format(
+          msg.frm.nick, emots[0]
+				)
```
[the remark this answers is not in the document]
sladyn98: Kindly use VSCode or any other editor to avoid such warnings.

tests/no_sir_test.py
```diff
@@ -0,0 +1,11 @@
+import os
+import no_sir
+from errbot.backends.test import testbot
+
+
+class Test_no_sir(object):
+    extra_plugin_dir = '.'
+
+    def test_no_sir(self, testbot):
+        testbot.push_message('sir')
```
sladyn98: Please read the above comment,this is just covering one of the scenarios.

tests/no_sir_test.py
```diff
@@ -0,0 +1,11 @@
...
+        assert 'Do not use sir in your conversation' in testbot.pop_message()
```
sladyn98: The assert statement does not match the bot output message, the test would always throw an assertion error because of the missing `emot` at the end

plugins/no_sir.py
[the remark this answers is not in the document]
AkshJain99: I use VS code only

plugins/no_sir.py
```diff
@@ -0,0 +1,20 @@
...
+    Do not use sir
+    """
+
+    def callback_message(self, msg):
+        emots = [':D']
+        match_sir = re.search(r'\bsir\b', '\bSir\b', msg.body)
+        if match_sir:
+            self.send(
+              msg.frm,
+              '@{}, Do not use sir in your conversation. {}'.format(
```
KVGarg: Please remove `comma` after mentioning the user!
Although the returned message is OK, I guess it could be more informative.
What do you think @sladyn98?

AkshJain99:
Okay ,sure I will do that in bit

plugins/no_sir.py
sladyn98: @KVGarg Definitely if you have any suggestions you could have written here :)

KVGarg: I guess we can frame a message somewhat like this
```
Hi @<username> :wave:! We don't use "Sir" in the community during the conversation. It's perfectly fine to call them via there usernames or handle. Also, If you to do such think you can better go with buddy :smiley: 
```
Something like this the message can be framed!

AkshJain99:
@kvgarg sure I will look after ur suggestions , actually I m quite busy for
a day or two soon I will look towards this pr
Thanks

tests/no_sir_test.py
meetmangukiya: @sladyn98  it's `in`

plugins/no_sir.py
```diff
@@ -0,0 +1,20 @@
+import re
+from errbot import BotPlugin
+
+
+class no_sir(BotPlugin):
```
meetmangukiya: class name not starting with caps :scream:

AkshJain99: Thats by mistake @meetmangukiya i will correct all the issues now in this pr

tests/no_sir_test.py
sladyn98: @meetmangukiya  Yeah your right,so it would pass, but it would be better to keep the outputs same although it doesnt matter :)

Issue #365, opened 2017-11-24
Default COBOT_PREFIX should not be the same as the production value


jayvdb:
Currently if someone starts the bot, and connects it to a server with a corobo production instance, two bots will respond to any `corobo ...` command, which will get very ugly when there are 20 or 30 kids trying it, allowing kids to play tricks on each other.

jayvdb:
Likewise multiple bots will be responding to `cobot ..` currently, which is best solved by removing that default configuration entirely: https://github.com/coala/corobo/issues/366

Makman2:
Tbh I don't consider this as critical. We can avoid the accidental clashing by forcing the user to pass a custom name for the bot. However, I haven't seen that yet so this sounds to me like an uncommon case. If people want to annoy people by spawning multiple instances that react on the same name, they could still. That's a misuse we can't stop expect individually banning people.

PR #652, opened 2019-03-18
config.py: Remove corobo as default prefix

sladyn98:
Removed `corobo` as default `BOT_PREFIX`
and added a check for empty `BOT_PREFIX`

Closes https://github.com/coala/corobo/issues/365

# Reviewers Checklist

### DIFF
--- a/config.py
+++ b/config.py
@@ -64,7 +64,10 @@ BOT_EXTRA_PLUGIN_DIR = BOT_ROOT
 BOT_LOG_FILE = os.path.join(BOT_ROOT, 'errbot.log')
 BOT_LOG_LEVEL = logging.DEBUG
 
-BOT_PREFIX = os.environ.get('BOT_PREFIX', 'corobo ')
+if not os.environ.get('BOT_PREFIX'):
+    raise SystemExit("Environment variable BOT_PREFIX not specified")
+
+BOT_PREFIX = os.environ.get('BOT_PREFIX')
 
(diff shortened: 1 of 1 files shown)


config.py
```diff
@@ -64,7 +64,10 @@
 BOT_LOG_FILE = os.path.join(BOT_ROOT, 'errbot.log')
 BOT_LOG_LEVEL = logging.DEBUG
 
-BOT_PREFIX = os.environ.get('BOT_PREFIX', 'corobo ')
+if not os.environ.get('BOT_PREFIX'):
+    raise SystemExit("Kindly specify BOT_PREFIX")
```
Vamshi99: `Kindly specify BOT_PREFIX` --> `Environment variable BOT_PREFIX not specified`

sladyn98:
@Vamshi99 made the required changes :)

sladyn98:
@KVGarg IMHO think we should leave those PR's separate since they are dealing with completely different issues. also our aim would be to get the build green upstream first and then deal with these PR's later.
Also we dont have gitmate now :( So these rebases will have to be done manually sigh:/

KVGarg:
> @KVGarg IMHO think we should leave those PR's separate since they are dealing with completely different issues. also our aim would be to get the build green upstream first and then deal with these PR's later.
> Also we dont have gitmate now :( So these rebases will have to be done manually sigh:/

Ohh, OK 👍🏻

PR #658, opened 2019-04-16
labhub.py: Add duplicate issue title check

sladyn98:
Added a duplicate issue check to warn
the user of a duplicate issue title.

Closes https://github.com/coala/corobo/issues/291

# Reviewers Checklist

### DIFF
--- a/plugins/labhub.py
+++ b/plugins/labhub.py
@@ -213,6 +213,14 @@ class LabHub(DefaultConfigMixin, BotPlugin):
 
         if repo_name in self.REPOS:
             repo = self.REPOS[repo_name]
+            query = iss_title + " is:issue " + "coala in:name"
+            issue_list = list(GitHub.raw_search(
+                GitHubToken(self.config['GH_TOKEN'], query)))
+            for issue in issue_list:
+                if iss_title.lower() == issue.title.lower():
+                    yield 'An issue with same title already exists at {}'.format(
+                        issue.web_url)
+                    return
             iss = repo.create_issue(iss_title, iss_description + extra_msg)
             yield 'Here you go: {}'.format(iss.web_url)
         else:
--- a/tests/labhub_test.py
+++ b/tests/labhub_test.py
@@ -199,6 +199,14 @@ class TestLabHub(LabHubTestCase):
             'You need to be a member of this organization to use this command.'
         )
 
+        # issue title already exists
+        mock_iss = create_autospec(IGitt.GitHub.GitHubIssue)
+        mock_iss.title = PropertyMock()
+        mock_iss.title = ('title',)
+        testbot_public.assertCommand(
+            '!!new issue coala title',
+            'An issue with same title already exists')
+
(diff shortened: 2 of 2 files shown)


abhishalya:
@sladyn98 Tests?

sladyn98:
@abhishalya @Vamshi99 I added some tests I am not very sure if they are correct, could you give me a review

plugins/labhub.py
```diff
@@ -213,6 +213,12 @@ def create_issue_cmd(self, msg, match):
 
         if repo_name in self.REPOS:
             repo = self.REPOS[repo_name]
+            issue_list = list(repo.search_issues())
```
Vamshi99: There will be thousands of issues in each repository, get the right ones and then check the body, title in those issues.

sladyn98: @Vamshi99 Im not quite sure I understand what your saying, how do I get the "right ones", I would have to scan all the issues in order to compare the titles, else how would I determine the duplicate one .

abhishalya: @sladyn98 You should have a look at the comments on the previous PR https://github.com/coala/corobo/pull/332 to get the correct approach of doing this.

sladyn98: @abhishalya Yeah I did have a look at #332 . For every ```assign_cmd``` the corresponding repo would be selected and iterated through to determine a duplicate issue list.
Another approach I see would be to maintain a cache of issues and their titles, but IMO that would be a unnecessary overhead because of the constant updation.

Vamshi99: @sladyn98 Use a search query, it will give the issues which matches the text you gave, then search in those issues. See GitHub search and IGitt API if you don't know how.

plugins/labhub.py
```diff
@@ -213,6 +213,14 @@ def create_issue_cmd(self, msg, match):
 
         if repo_name in self.REPOS:
             repo = self.REPOS[repo_name]
+            query = iss_title + " is:issue"
```
Vamshi99: How does this query help?  You need to search for `open` issues of `coala` repositories.

Vamshi99: Also, I couldn't find any implementation for GitLab. Search for appropriate method in IGitt API before you do anything.

sladyn98: @Vamshi99 Yeah still working on it.Just wanted you to review the approach.Thanks

sladyn98: > How does this query help? You need to search for `open` issues of `coala` repositories.

@Vamshi99 this would help us get all the issues that have been openened that contain the current issue title open or close.Because it could be that the issue we are creating has already been solved, morevoer it is no possible to include repo in the query so this would list out all issues with out title.

sladyn98:
> Possible to make an exact title search to avoid too many results?

Actually the title search is being done at `iss_title`  this and the other constraints do allow us to minimize search results

sladyn98:
> Not necessary to split the string further ;)

Where exactly ?

